PR #11, opened 2018-04-07
 Fix mentioning of renamed ReportingMode in user guide

BlueIce:
---

I hereby agree to the terms of the jqwik Contributor Agreement.


jlink:
@BlueIce Good catch!

Since it's the first external pull request I had to come up with a contributor's agreement ad hoc.

Please take a look at it here: https://github.com/jlink/jqwik/blob/master/CONTRIBUTING.md

Then add 
```
---
I hereby agree to the terms of the jqwik Contributor Agreement.
```

To the description of your PR. Thanks!

BlueIce:
@jlink: I've updated the description.

Issue #55, opened 2019-04-11
JRE dependent "whitespace" generation


tmohme:
## Testing Problem

Unfortunately what is considered as whitespace character (`Character.isWhitespace(...)`) has changed between Java 8 (OpenJDK) and Java 11 (OpenJDK):  
In particular the U+180E ([Mongolian Vovel Separator](https://unicode.org/cldr/utility/character.jsp?a=180e&B1=Show)) counts in Java 8 as whitespace, while it does not in Java 11.  

This causes problems because when generating whitespace chars with jqwik (which contains hard-coded lists of whitespace chars based on Java 8) and actually running the functionality on Java 11.

## Suggested Solution

Instead of using a prepared, hard-coded list of whitespace chars, the chars considered as whitespace by the actual runtime should be used.  
Please have a look at my corresponding pull request.

## Discussion

To minimize the runtime overhead, the determination of the whitespace chars could (should) be done at class-level instead of instance-level.

jlink:
Implemented with https://github.com/jlink/jqwik/commit/64cfac057e0df2cd6be82c65878825a71c4c5c30

jlink:
Available in 1.1.3-SNAPSHOT

jlink:
Released in 1.1.3

PR #56, opened 2019-04-11
make whitespace determination dynamic

tmohme:
## Overview

Like described in [issue #55](https://github.com/jlink/jqwik/issues/55), using the Java 8 fixed set of whitespace characters leads to wrong results in a Java 11 environment.  

This change aims to determine the set of whitespace characters at runtime, thus avoiding problems caused by changes between Java versions.

---

### DIFF
--- a/engine/src/main/java/net/jqwik/engine/properties/arbitraries/DefaultStringArbitrary.java
+++ b/engine/src/main/java/net/jqwik/engine/properties/arbitraries/DefaultStringArbitrary.java
@@ -9,6 +9,19 @@ import net.jqwik.engine.properties.arbitraries.exhaustive.*;
 import net.jqwik.engine.properties.arbitraries.randomized.*;
 
 public class DefaultStringArbitrary extends AbstractArbitraryBase implements StringArbitrary {
+	public static final char[] WHITESPACE_CHARS;
+
+	static {
+		// determine WHITESPACE_CHARS at runtime because the environments differ . . .
+		final StringBuilder whitespace = IntStream.range(Character.MIN_VALUE, Character.MAX_VALUE + 1)
+												  .filter(Character::isWhitespace)
+												  .collect(StringBuilder::new, StringBuilder::appendCodePoint, StringBuilder::append);
+
+		final int whitespaceLength = whitespace.length();
+		final char[] charArray = new char[whitespaceLength];
+		whitespace.getChars(0, whitespaceLength, charArray, 0);
+		WHITESPACE_CHARS = charArray;
+	}
 
 	private CharacterArbitrary characterArbitrary = new DefaultCharacterArbitrary();
 
@@ -82,46 +95,9 @@ public class DefaultStringArbitrary extends AbstractArbitraryBase implements Str
 		return clone;
 	}
 
-	/**
-	 * Extracted unicodes from java 8 with
-	 * <pre>
-	 * 	for (char c = Character.MIN_VALUE; c &lt; Character.MAX_VALUE; c++) {
-	 * 		if (Character.isWhitespace(c)) {
-	 * 			System.out.println( "\\u" + Integer.toHexString(c | 0x10000).substring(1) );
-	 * 		}
-	 * 	}
-	 *  </pre>
-	 */
 	@Override
 	public StringArbitrary whitespace() {
-		return this.withChars( //
-			'\u0009', //
-			'\n', //
-			'\u000b', //
-			'\u000c', //
-			'\r', //
-			'\u001c', //
-			'\u001d', //
-			'\u001e', //
-			'\u001f', //
-			'\u0020', //
-			'\u1680', //
-			'\u180e', //
-			'\u2000', //
-			'\u2001', //
-			'\u2002', //
-			'\u2003', //
-			'\u2004', //
-			'\u2005', //
-			'\u2006', //
-			'\u2008', //
-			'\u2009', //
-			'\u200a', //
-			'\u2028', //
-			'\u2029', //
-			'\u205f', //
-			'\u3000' //
-		);
+		return this.withChars(WHITESPACE_CHARS);
 	}
 
 	@Override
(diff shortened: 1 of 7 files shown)


engine/src/main/java/net/jqwik/engine/properties/configurators/WhitespaceConfigurator.java
```diff
@@ -0,0 +1,25 @@
...
+
+	public StringArbitrary configure(StringArbitrary arbitrary, Whitespace whitespace) {
+		return arbitrary.withChars(DefaultStringArbitrary.WHITESPACE_CHARS);
+	}
+
+	public CharacterArbitrary configure(CharacterArbitrary arbitrary, Whitespace whitespace) {
+		CharacterArbitrary result = arbitrary;
+
+		for (char c : DefaultStringArbitrary.WHITESPACE_CHARS) {
+			result = result.between(c, c);
```
jlink: Problem is that `DefaultCharacterArbitrary.between()` does not add the range but replaces a previous range :-( To accomplish what you want will probably require a change to `DefaultCharacterArbitrary`.

engine/src/test/java/net/jqwik/api/constraints/WhitespaceProperties.java
```diff
@@ -0,0 +1,24 @@
...
+
+@Group
+class WhitespaceProperties {
+
+	@Property
+	boolean stringWithWhitespace(@ForAll @Whitespace String aString) {
+		return aString.chars().allMatch(Character::isWhitespace);
+	}
+
+	@Property
```
jlink: The property is a bit too weak, since it does not check that all whitespace characters will be generated at some time. Such properties would cover the problem mentioned above.

engine/src/main/java/net/jqwik/engine/properties/configurators/WhitespaceConfigurator.java
```diff
@@ -0,0 +1,25 @@
...
+import net.jqwik.api.arbitraries.CharacterArbitrary;
+import net.jqwik.api.arbitraries.StringArbitrary;
+import net.jqwik.api.configurators.ArbitraryConfiguratorBase;
+import net.jqwik.api.constraints.Whitespace;
+import net.jqwik.engine.properties.arbitraries.DefaultStringArbitrary;
+
+public class WhitespaceConfigurator extends ArbitraryConfiguratorBase {
+
+	public StringArbitrary configure(StringArbitrary arbitrary, Whitespace whitespace) {
+		return arbitrary.withChars(DefaultStringArbitrary.WHITESPACE_CHARS);
```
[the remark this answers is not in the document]
tmohme: done

engine/src/main/java/net/jqwik/engine/properties/arbitraries/DefaultStringArbitrary.java
```diff
@@ -9,6 +9,19 @@
 import net.jqwik.engine.properties.arbitraries.randomized.*;
 
 public class DefaultStringArbitrary extends AbstractArbitraryBase implements StringArbitrary {
+	public static final char[] WHITESPACE_CHARS;
+
+	static {
```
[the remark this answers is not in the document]
tmohme: OK for me, but . . .
* Still on class-level (as opposed to instance level)?!
* Any preferences with regard to concurrency?  
  As the computed value is immutable and will always be the same (within one run) there's effectively no need for synchronization. The worst thing that could happen is multiple threads compute the same value.

engine/src/test/java/net/jqwik/api/constraints/WhitespaceProperties.java
tmohme: I already thought of this but didn't have a catchy idea how to achieve it with reasonable effort.  
Any suggestions?

engine/src/main/java/net/jqwik/engine/properties/configurators/WhitespaceConfigurator.java
tmohme: `DefaultCharacterArbitrary` doesn't look too complicated, but I see several design-decisions coming-up that I don't want to do with comment-ping-pong.  
Could you describe in advance the design goals you have in mind?  
Or should we simply go without this enhancement and instead make the documentation more precise (e.g. point out that only one (the last?) of multiple of the constraint annotations will be respected when generating chars)?

jlink: I updated DefaultCharacterArbitrary to add up chosen ranges and enhanced it with the capability to allow a set of chars. see `DefaultCharacterArbitrary.with(chars... allowedChars)`. This should suit the purpose I hope.

engine/src/test/java/net/jqwik/api/constraints/WhitespaceProperties.java
jlink: `WhitespaceProperties` is fine I think. You should consider to add tests to `DefaultStringArbitraryTests` and `DefaultCharacterArbitraryTests`. The examples there do similar things.

jlink:
Please rebase with master to get the latest changes in `CharacterArbitrary` and `DefaultCharacterArbitrary`. Thanks for the effort!

engine/src/test/java/net/jqwik/api/constraints/WhitespaceProperties.java
tmohme: Added tests.

tmohme:
I guess you want to take care of the lazy init thing yourself?  
Besides this, I think my initial choice to place the `WHITESPACE_CHARS` in `DefaultStringArbitrary` was not optimal. Meanwhile I would place it in `DefaultCharacterArbitrary` because this is one of the building-blocks the `DefaultStringArbitrary` is made of and it should be able to provide the capability by itself, without relying on `DefaultStringArbitrary`.  
A dependency from `DefaultStringArbitrary` to `DefaultCharacterArbitrary` is imho OK.

jlink:
> I guess you want to take care of the lazy init thing yourself?

OK, I can do that. Thanks again!

PR #63, opened 2019-06-20
allow non-static `DomainContext`s to be defined in base classes

larsrh:
## Overview

When using the `@Domain` annotation with inner classes, instantiation only works on `static` inner classes or non-`static` inner classes in the same class. There is no technical reason why non-`static` inner classes inherited from a superclass shouldn't work.

### Details

I changed a `Class<?>` _equality_ check to an `isAssignableFrom` check.

Also, we already spent time figuring out the `outerClass` ourselves, so we can also compute the `Constructor<?>` ourselves. This is necessary for the following reason:

– Previously, the code called `ReflectionSupport.newInstance(clazz, parentInstance)` from JUnit.
– JUnit will look at the class of `parentInstance` and will try to figure out the correct constructor in `clazz`.
– However, if `parentInstance` is a subclass of `outerClass` (which this PR enables), then there is no appropriate constructor.
– Consequently, we need to select the constructor "statically" from the `clazz`.

(I've tried this out, and without that second change, my test case fails).

---

### DIFF
--- a/engine/src/test/java/net/jqwik/api/BaseDomainContexts.java
+++ b/engine/src/test/java/net/jqwik/api/BaseDomainContexts.java
@@ -1,0 +1,15 @@
+package net.jqwik.api;
+
+import net.jqwik.api.domains.AbstractDomainContextBase;
+
+abstract class BaseDomainContexts {
+
+	Integer[] specificNumbers;
+
+	class SpecificNumbersContext extends AbstractDomainContextBase {
+		private SpecificNumbersContext() {
+			registerArbitrary(Integer.class, Arbitraries.of(specificNumbers));
+		}
+	}
+
+}
--- a/engine/src/main/java/net/jqwik/engine/support/JqwikReflectionSupport.java
+++ b/engine/src/main/java/net/jqwik/engine/support/JqwikReflectionSupport.java
@@ -13,6 +13,8 @@ import org.junit.platform.commons.support.*;
 import net.jqwik.api.lifecycle.*;
 import net.jqwik.api.providers.*;
 import net.jqwik.engine.discovery.predicates.*;
+import org.junit.platform.commons.util.ExceptionUtils;
+import org.junit.platform.commons.util.ReflectionUtils;
 
 import static java.util.stream.Collectors.*;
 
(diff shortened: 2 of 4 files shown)


engine/src/test/java/net/jqwik/api/BaseDomainContexts.java
```diff
@@ -0,0 +1,15 @@
+package net.jqwik.api;
+
+import net.jqwik.api.domains.AbstractDomainContextBase;
+
+abstract class BaseDomainContexts {
+
+	Integer[] specificNumbers;
```
larsrh: I should note that I put this in there to ensure that the inner class is non-`static`.

jlink:
Thanks for the PR. Enabling functionality that should work out of symmetry reasons is good!

There might also be hidden another test (readability) problem underneath. How did you stumble upon this issue?

jlink:
If you're ambitious you can add a group of tests for `newInstanceInTestContext(...)` in `JqwikReflectionSupportTests` similiar to `@Group class NewInstanceWithDefaultConstructor`.

larsrh:
> There might also be hidden another test (readability) problem underneath. How did you stumble upon this issue?

I'm currently working on a TCK for an API. Very simply speaking, implementers of that API need to provide some domain classes and some constructors for such classes. The API defines the shape of these constructors. Consequently, I can also define some `Arbitrary` instances for them.

My plan is to provide some `DomainContext` instances for those API classes in the TCK. However, the domain classes are scoped to the API. I cannot create them without also knowing the API.

Trimmed down, this would look as follows:

```java
public abstract class TCKBase {

  // parametrized over the API
  private final API impl;

  protected TCK(API impl) {
    this.impl = impl;
  }

  protected APIContext extends AbstractDomainContextBase {
    public APIContext() {
      registerArbitrary(...) // using TCKBase.this.impl
    }
  }

}
```

... and in the tests:

```java
public class TCKFooTest extends TCKBase {

  public TCKFooTest() {
    // get an impl from somewhere ...
  }

  @Property @Domain(APIContext.class)
  public void testDomainType(API.DomainType bla) {
    // test something
  }

}
```

I implemented that with an inner class because that is much simpler and requires less machinery than somehow adding support for a DI framework.

larsrh:
> If you're ambitious you can add a group of tests

Working on it.

larsrh:
Added more low-level tests for this.

jlink:
Thanks again. TCKs are an interesting use case!

jlink:
BTW, I had to get rid of access to junit platform's ExceptionUtils and ReflectionUtils since they are not part of JUnit's public API.

larsrh:
Thanks for the quick merging! Would it be possible to get a new release containing this change?

jlink:
Already available in version 1.1.6-SNAPSHOT in case you need it.

jlink:
There's one missing feature for 1.1.6. I hope to have it ready next weekend.

jlink:
1.1.6 has been pushed to maven central. Should show up in the next hour.

larsrh:
Thanks a lot for the quick turnaround!

PR #66, opened 2019-07-24
Fix Unshrinkable null value equals And hashCode error

mhyeon-lee:
## Overview

- Unshrinkable null value got exception when call equals or hashCode method

### Details

- Fix equals and hashCode method to escape NPE.

---

I hereby agree to the terms of the jqwik Contributor Agreement.

### DIFF
--- a/engine/src/test/java/net/jqwik/engine/properties/shrinking/UnshrinkableTests.java
+++ b/engine/src/test/java/net/jqwik/engine/properties/shrinking/UnshrinkableTests.java
@@ -1,0 +1,31 @@
+package net.jqwik.engine.properties.shrinking;
+
+import net.jqwik.api.*;
+import net.jqwik.api.arbitraries.SizableArbitrary;
+import org.assertj.core.api.Assertions;
+
+import java.util.Random;
+import java.util.Set;
+
+class UnshrinkableTests {
+	@Example
+	void nullValueEquals() {
+		Unshrinkable<?> unshrinkable1 = new Unshrinkable<>(null);
+		Unshrinkable<?> unshrinkable2 = new Unshrinkable<>(null);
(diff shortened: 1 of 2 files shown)


jlink:
Thanks for the catch! Very valuable. I have two minor change requests. See there...

engine/src/test/java/net/jqwik/engine/properties/shrinking/UnshrinkableTests.java
```diff
@@ -0,0 +1,33 @@
+package net.jqwik.engine.properties.shrinking;
+
+import net.jqwik.api.Arbitraries;
+import net.jqwik.api.Example;
```
jlink: jqwik coding style is to use only wildcard imports, e.g. `import net.jqwik.api.*`

engine/src/test/java/net/jqwik/engine/properties/shrinking/UnshrinkableTests.java
```diff
@@ -0,0 +1,33 @@
...
+	}
+
+	@Example
+	void nullValueHashCode() {
+		Unshrinkable<?> unshrinkable = new Unshrinkable<>(null);
+		Assertions.assertThat(unshrinkable.hashCode()).isEqualTo(0);
+	}
+
+	@Property(tries = 50)
+	void nullValueUnshrinkable() {
```
jlink: In properties you should never use `new Random()` directly because this will prevent deterministic repeats. Instead use a parameter `@ForAll Random random` and use that.

mhyeon-lee: @jlink
Thank you for your review. I have amended your request.

jlink:
Thanks again. Welcome to jqwik's list of contributors.

jlink:
Available in 1.2.0-SNAPSHOT

PR #78, opened 2019-12-15
fix wrong link

luvarqpp:
Link for example was not correct. Now it points to https://github.com/jlink/jqwik/blob/1.1.2/documentation/src/test/java/net/jqwik/docs/types/TypeArbitraryExamples.java instead of https://github.com/jlink/jqwik/blob/1.2.1/documentation/src/test/java/net/jqwik/docs/types/PersonProperties.java .

## Overview

_Formulate the problem / feature your pull request is aiming at_

### Details

_Give any necessary details about your code that is not obvious from the code itself._

_List open issues and remaining problems with your solution._

---

### DIFF
--- a/docs/docs/1.1.0/user-guide.md
+++ b/docs/docs/1.1.0/user-guide.md
@@ -2385,7 +2385,7 @@ the class in order to generate instances. Whenever there's an exception during
 generation they will be ignored; that way you'll only get valid instances.
 
 There are quite a few ways usage and configuration options. Have a look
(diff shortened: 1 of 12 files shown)


jlink:
Many thanks for the catch. 

Would you be willing to fix it in all user-guide versions (1.1.0 - 1.1.6, 1.2.0 - 1.2.1)  and most importantly in the template file: https://github.com/jlink/jqwik/blob/master/documentation/src/docs/user-guide.template.md?

luvarqpp:
Ehm, I was a bit confused when I have checkout branch 1.1.0 and than I have realized, that all othr versions are in master branch :)

Fixed all and also current and snapshot files (grep through all of files).

jlink:
Many thanks!

PR #117, opened 2020-08-10
initilize constants thread safe

Danny02:
## Overview

I want to use Jqwik Arbitraries in normal JUnit test to generate some example data. I get errors when I run those test in parallel, because Jqwik does not initialize constants (e.g. Provider) in a thread safe way.

---

### DIFF
--- a/engine/src/main/java/net/jqwik/engine/LazyServiceLoaderCache.java
+++ b/engine/src/main/java/net/jqwik/engine/LazyServiceLoaderCache.java
@@ -1,0 +1,26 @@
+package net.jqwik.engine;
+
+import java.util.List;
+import java.util.ServiceLoader;
+import java.util.concurrent.CopyOnWriteArrayList;
+
+public class LazyServiceLoaderCache<S> {
+	private final Class<S> clz;
+	private List<S> services;
+
+	public LazyServiceLoaderCache(Class<S> clz) {
+		this.clz = clz;
+	}
+
+	public List<S> getServices() {
+		if (services == null) {
+			loadServices();
+		}
+		return services;
+	}
+
+	private synchronized void loadServices() {
+		services = new CopyOnWriteArrayList<>();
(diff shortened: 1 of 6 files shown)


jlink:
Hi @Danny02! thanks for the PR. I obviously did not think about parallel execution when implementing most of the static initializers. :-(

However, I always try to steer clear off static initializer blocks since one has no control over when they are being executed - and initialization may be costly.

Thus, I'd prefer to make the accessor methods `synchronized` as it already is in class `RegisteredLifecycleHooks`. Would you be willing to make that change?

Danny02:
Hi,

in the end it is of course your decision, but I'm under the impression that in this case static initializers may be the better solution. (I may have forgotten to make the fields final)

* static initializers are run the first time the class is used (in the JVM at least). In the case of those classes, the runtime behavior will not change, because the only functionality they have are the collection getters (which are the first interaction with those classes). (see [Effective Java, Joshua Bloch](https://www.kea.nu/files/textbooks/new/Effective%20Java%20%282017%2C%20Addison-Wesley%29.pdf) p. 334)
* they are thread-safe because the JVM guarantees, that they are called only once and before any further interaction with the class
* synchronized introduces a probably neglectable runtime cost

jlink:
Yes, I've re-read about static initializers and their pros and cons. 

Performance will probably not change with normal jqwik usage but when running individual jqwik tests since many do not need all the loaded services. But the real reason is my dislike for losing control over the exact time static initializers are executed.

So my current call is going for synchronized. Which does not exclude changing my mind sometime in the future.

Danny02:
ok, will do the changes

jlink:
Thanks!

jlink:
Looks very good. Thank you!

jlink:
Just published "1.3.4-SNAPSHOT" with this PR included

jlink:
Is now available in release 1.3.4

PR #128, opened 2020-10-27
E-Mail Addresses

zinki97:
Generates E-Mail Addresses

### DIFF
--- a/api/src/main/java/net/jqwik/api/Arbitraries.java
+++ b/api/src/main/java/net/jqwik/api/Arbitraries.java
@@ -1,5 +1,6 @@
 package net.jqwik.api;
 
+import java.net.*;
 import java.util.*;
 import java.util.function.*;
 import java.util.stream.*;
@@ -456,7 +457,27 @@ public class Arbitraries {
 	}
 
 	private static Arbitrary<String> emailsDomain(){
-		Arbitrary<String> domain = Arbitraries.strings().alpha().ofLength(10);
+		return Arbitraries.frequencyOf(
+				Tuple.of(1, emailsDomainIPv4()),
+				Tuple.of(1, emailsDomainIPv6()),
+				Tuple.of(2, emailsDomainDomain())
+		);
+	}
+
+	private static Arbitrary<String> emailsDomainIPv4(){
+		Arbitrary<Integer> addressPart = Arbitraries.integers().between(0, 255);
+		return Combinators.combine(addressPart, addressPart, addressPart, addressPart).as((a, b, c, d) -> "[" + a + "." + b + "." + c + "." + d + "]");
+	}
+
+	private static Arbitrary<String> emailsDomainIPv6(){
+		//Arbitrary<String> addressPart = Arbitraries.strings().numeric().withChars('a', 'b', 'c', 'd', 'e', 'f', 'A', 'B', 'C', 'D', 'E', 'F').ofMinLength(1).ofMaxLength(4);
+		//Arbitrary<String> address = Combinators.combine(addressPart, addressPart, addressPart, addressPart, addressPart, addressPart, addressPart, addressPart).as((a, b, c, d, e, f, g, h) -> "[" + a + ":" + b + ":" + c + ":" + d + ":" + e + ":" + f + ":" + g + ":" + h + "]");
+		Arbitrary<String> address = Arbitraries.strings().numeric().alpha().ofLength(10);
+		return address;
+	}
+
+	private static Arbitrary<String> emailsDomainDomain(){
+		Arbitrary<String> domain = Arbitraries.strings().numeric().alpha().ofLength(10);
 		return domain;
 	}
 
--- a/api/src/main/java/net/jqwik/api/Arbitraries.java
+++ b/api/src/main/java/net/jqwik/api/Arbitraries.java
@@ -428,8 +428,8 @@ public class Arbitraries {
 	 */
 	@API(status = EXPERIMENTAL, since = "1.38")
 	public static Arbitrary<String> emails() {
-		//return just("test@test.b.com");
-		return just("test!!@[192.21.2.1]");
+		return just("test@test.b.com");
+		//return just("test@[192.21.2.1]");
 	}
 
 	/**
--- a/api/src/main/java/net/jqwik/api/Arbitraries.java
+++ b/api/src/main/java/net/jqwik/api/Arbitraries.java
@@ -430,7 +430,17 @@ public class Arbitraries {
 	@API(status = EXPERIMENTAL, since = "1.38")
 	public static Arbitrary<String> emails() {
 		//return just("test@test.b.com");
-		return just("test@[192.21.2.1]");
+		Arbitrary<String> arbitraryFirst = emailsFirstPart();
+		StringArbitrary arbitraryAt = Arbitraries.strings().withChars('@').ofLength(1);
+		return Combinators.combine(arbitraryFirst, arbitraryAt).as((partA, at) -> partA + at);
+	}
+
+	private static Arbitrary<String> emailsFirstPart(){
+		Arbitrary<String> unquoted = Arbitraries.strings().alpha().numeric().withChars("!#$%&'*+-/=?^_`{|}~.").ofMinLength(1).ofMaxLength(64);
+		unquoted = unquoted.filter(v -> !v.contains(".."));
+		unquoted = unquoted.filter(v -> v.charAt(0) != '.');
+		unquoted = unquoted.filter(v -> v.charAt(v.length() - 1) != '.');
+		return unquoted;
 	}
 
 	/**
--- a/api/src/main/java/net/jqwik/api/Arbitraries.java
+++ b/api/src/main/java/net/jqwik/api/Arbitraries.java
@@ -1,6 +1,5 @@
 package net.jqwik.api;
 
-import java.net.*;
 import java.util.*;
 import java.util.function.*;
 import java.util.stream.*;
@@ -70,6 +69,8 @@ public class Arbitraries {
 		public abstract ShortArbitrary shorts();
 
 		public abstract StringArbitrary strings();
+
+		public abstract Arbitrary<String> emails();
 
 		public abstract CharacterArbitrary chars();
 
@@ -427,58 +428,9 @@ public class Arbitraries {
 	 *
 	 * @return a new arbitrary instance
 	 */
-	public static int i = 0;
 	@API(status = EXPERIMENTAL, since = "1.38")
 	public static Arbitrary<String> emails() {
-		Arbitrary<String> arbitraryLocalPart = emailsLocalPart();
-		Arbitrary<String> arbitraryDomain = emailsDomain();
-		return Combinators.combine(arbitraryLocalPart, arbitraryDomain).as((localPart, domain) -> localPart + "@" + domain);
-	}
-
-	private static Arbitrary<String> emailsLocalPart(){
-		Arbitrary<String> unquoted = emailsLocalPartUnquoted();
-		Arbitrary<String> quoted = emailsLocalPartQuoted();
-		return oneOf(unquoted, quoted);
-	}
-
-	private static Arbitrary<String> emailsLocalPartUnquoted(){
-		Arbitrary<String> unquoted = Arbitraries.strings().alpha().numeric().withChars("!#$%&'*+-/=?^_`{|}~.").ofMinLength(1).ofMaxLength(64);
-		unquoted = unquoted.filter(v -> !v.contains(".."));
-		unquoted = unquoted.filter(v -> v.charAt(0) != '.');
-		unquoted = unquoted.filter(v -> v.charAt(v.length() - 1) != '.');
-		return unquoted;
-	}
-
-	private static Arbitrary<String> emailsLocalPartQuoted(){
-		Arbitrary<String> quoted = Arbitraries.strings().alpha().numeric().withChars(" !#$%&'*+-/=?^_`{|}~.\"(),:;<>@[\\]").ofMinLength(1).ofMaxLength(62);
-		quoted = quoted.map(v -> "\"" + v.replace("\\", "\\\\").replace("\"", "\\\"") + "\"");
-		quoted = quoted.filter(v -> v.length() <= 64);
-		return quoted;
-	}
-
-	private static Arbitrary<String> emailsDomain(){
-		return Arbitraries.frequencyOf(
-				Tuple.of(1, emailsDomainIPv4()),
-				Tuple.of(1, emailsDomainIPv6()),
-				Tuple.of(2, emailsDomainDomain())
-		);
-	}
-
-	private static Arbitrary<String> emailsDomainIPv4(){
-		Arbitrary<Integer> addressPart = Arbitraries.integers().between(0, 255);
-		return Combinators.combine(addressPart, addressPart, addressPart, addressPart).as((a, b, c, d) -> "[" + a + "." + b + "." + c + "." + d + "]");
-	}
-
-	private static Arbitrary<String> emailsDomainIPv6(){
-		//Arbitrary<String> addressPart = Arbitraries.strings().numeric().withChars('a', 'b', 'c', 'd', 'e', 'f', 'A', 'B', 'C', 'D', 'E', 'F').ofMinLength(1).ofMaxLength(4);
-		//Arbitrary<String> address = Combinators.combine(addressPart, addressPart, addressPart, addressPart, addressPart, addressPart, addressPart, addressPart).as((a, b, c, d, e, f, g, h) -> "[" + a + ":" + b + ":" + c + ":" + d + ":" + e + ":" + f + ":" + g + ":" + h + "]");
-		Arbitrary<String> address = Arbitraries.strings().numeric().alpha().ofLength(10);
-		return address;
-	}
-
-	private static Arbitrary<String> emailsDomainDomain(){
-		Arbitrary<String> domain = Arbitraries.strings().numeric().alpha().ofLength(10);
-		return domain;
+		return ArbitrariesFacade.implementation.emails();
 	}
 
 	/**
--- a/api/src/main/java/net/jqwik/api/Arbitraries.java
+++ b/api/src/main/java/net/jqwik/api/Arbitraries.java
@@ -429,7 +429,6 @@ public class Arbitraries {
 	public static int i = 0;
 	@API(status = EXPERIMENTAL, since = "1.38")
 	public static Arbitrary<String> emails() {
-		//return just("test@a.test.com");
 		Arbitrary<String> arbitraryLocalPart = emailsLocalPart();
 		Arbitrary<String> arbitraryDomain = emailsDomain();
 		return Combinators.combine(arbitraryLocalPart, arbitraryDomain).as((localPart, domain) -> localPart + "@" + domain);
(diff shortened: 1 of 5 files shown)


zinki97:
Includes first testcases for E-Mail Addresses:
- Contains "@"
- Valid length of address
- Valid use of "." and "-" (Top-Level-Domain min. 2 signs)
- Only valid Signs after the "@"
- Valid Domain-Length (even Subdomain-Length)
- Valid IP-Address (only if IP-Address is used after the "@")

zinki97:
Added the next (last?) testcases:
- Valid use of "."
- Only valid signs before the "@"

jlink:
I wonder if the checks in the tests could be simplified by partially using regex matches.

engine/src/test/java/net/jqwik/api/EMailsTests.java
```diff
@@ -0,0 +1,228 @@
+package net.jqwik.api;
+
+import com.tngtech.archunit.thirdparty.com.google.common.net.*;
+
+import static net.jqwik.api.ArbitraryTestHelper.*;
+
+public class EMailsTests {
```
jlink, line 7: Test classes need not be public. All (most) others in the jqwik codebase are not.

engine/src/test/java/net/jqwik/api/EMailsTests.java
```diff
@@ -0,0 +1,228 @@
...
+
+	@Example
+	void containsAtSign(){
+		RandomGenerator<String> generator = getEmailGeneratorFromArbitrary();
+		assertAllGenerated(generator, (String value) -> value.contains("@"));
+	}
```
jlink, line 13: Have you tried using a "real" property instead, like 

```
	@Property
	void containsAtSign(@ForAll("emails") String email) {
        Assertions.assertThat(email).contains("@");
	}
    @Provide
    Arbitrary<String> emails() {
        return Arbitraries.emails();
    }
```

zinki97: Only here? Or for all @Example ?

zinki97: > I wonder if the checks in the tests could be simplified by partially using regex matches.

Sure, but I like it more on this way. Is this a problem?

jlink: Whatever you think is more readable.

jlink: > Only here? Or for all `@Example ?`

Be consistent. I prefer properties whenever I can. In this case they seem a good fit to me.

zinki97:
Update: I've changed all `@Example` to `@Property`

zinki97:
Update: Added check for the statistic: `isQuoted` / `isIPAddress`

Is this correct?

engine/src/test/java/net/jqwik/api/EMailsTests.java
```diff
@@ -0,0 +1,208 @@
...
+		String secondPart = getSecondPartOfEmail(email);
+		if(!isIPAddress(secondPart)){
+			return;
+		}
+		secondPart = secondPart.substring(1, secondPart.length() - 1);
+		Assertions.assertThat(InetAddresses.isInetAddress(secondPart)).isTrue();
+	}
+
+	@Property
+	void checkFirstPartStatistic(@ForAll("emails") String email){
```
jlink, line 156: I wonder if the name of the method can be improved. what about `quotedAndUnquotedUsernamesAreGenerated`

engine/src/test/java/net/jqwik/api/EMailsTests.java
```diff
@@ -0,0 +1,208 @@
...
+	}
+
+	private boolean isQuoted(String firstPart){
+		if(firstPart.length() >= 3 && firstPart.charAt(0) == '"' && firstPart.charAt(firstPart.length() - 1) == '"'){
+			return true;
+		}
+		return false;
+	}
+
+	private String getFirstPartOfEmail(String email){
```
jlink, line 189: "first part" is called "local part" in the wikipedia article. Maybe that's a better name throughout the tests?

engine/src/test/java/net/jqwik/api/EMailsTests.java
```diff
@@ -0,0 +1,208 @@
...
+		int index = email.indexOf('@');
+		if(index == -1){
+			index = 0;
+		}
+		return email.substring(0, index);
+	}
+
+	private String getSecondPartOfEmail(String email){
```
jlink, line 197: Same for "second part": It's called "domain" in the article.

engine/src/test/java/net/jqwik/api/EMailsTests.java
```diff
@@ -0,0 +1,208 @@
...
+			if (secondPart.contains(".")) {
+				int startVal = 0;
+				int checkVal = secondPart.indexOf('.');
+				Assertions.assertThat(checkVal - startVal).isLessThanOrEqualTo(63);
+				secondPart = secondPart.substring(checkVal + 1);
+			} else {
+				Assertions.assertThat(secondPart.length()).isLessThanOrEqualTo(63);
+				return;
+			}
+		}
```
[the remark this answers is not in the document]
zinki97: The while loop checks whether each part of the domain has a maximum length of 63 characters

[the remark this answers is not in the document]
jlink: What about saying exactly that in the code? 
I see various options:
- Split domain in its parts and then check parts using a stream.
- Or you could build a special AssertJ matcher for doing that.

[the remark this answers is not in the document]
zinki97: Is it better now?

zinki97:
Update: LocalPart is now generated and I fixed a bug in `getLocalPartOfEmail` and `getDomainOfEmail`

zinki97:
I think I've found a better way than this. I'll change it tomorrow.

`quoted = quoted.filter(v -> !containsSignWithoutBackslash(v, '"'));
		quoted = quoted.filter(v -> everyBackslashHasBackslash(v));
		`

engine/src/test/java/net/jqwik/api/EMailsTests.java
```diff
@@ -0,0 +1,206 @@
...
+	}
+
+	@Property
+	void validLengthAfterAt(@ForAll("emails") String email){
+		String domain = getDomainOfEmail(email);
+		Assertions.assertThat(domain.length()).isBetween(1, 253);
+	}
+
+	@Property
+	void validSignsBeforeAt(@ForAll("emails") String email){
```
jlink, line 31: This property is way too long for my taste, I suggest two things: 
1) Make it two properties. One for quoted, one for unquoted local part.
2) Extract some meaningful assertion methods like "containsOnlyCharsAllowedInLocalPart" or such.
Moreover, removing valid characters until the string is supposed to be empty is clever but not readable IMO.

engine/src/test/java/net/jqwik/api/EMailsTests.java
```diff
@@ -0,0 +1,206 @@
...
+	void validUseOfDotBeforeAt(@ForAll("emails") String email){
+		String localPart = getLocalPartOfEmail(email);
+		Assume.that(!isQuoted(localPart));
+		Assertions.assertThat(localPart).doesNotContain("..");
+		Assertions.assertThat(localPart.charAt(0)).isNotEqualTo('.');
+		Assertions.assertThat(localPart.charAt(localPart.length() - 1)).isNotEqualTo('.');
+	}
+
+	@Property
+	void validSignsAfterAt(@ForAll("emails") String email){
```
jlink, line 91: See my comment for validSignsBeforeAt

api/src/main/java/net/jqwik/api/Arbitraries.java
```diff
@@ -421,6 +422,65 @@ public static StringArbitrary strings() {
 		return ArbitrariesFacade.implementation.strings();
 	}
 
+	/**
+	 * Create an arbitrary that generates valid E-Mail addresses.
+	 *
+	 * @return a new arbitrary instance
+	 */
+	public static int i = 0;
```
jlink, line 430: What is i being used for?

api/src/main/java/net/jqwik/api/Arbitraries.java
```diff
@@ -421,6 +422,65 @@ public static StringArbitrary strings() {
...
+	@API(status = EXPERIMENTAL, since = "1.38")
+	public static Arbitrary<String> emails() {
```
jlink, line 432: Implementation of emails() looks good to me. Please move it to ArbitrariesFacadeImple

engine/src/test/java/net/jqwik/api/EMailsTests.java
```diff
@@ -0,0 +1,204 @@
...
+
+import net.jqwik.api.statistics.*;
+
+import org.assertj.core.api.*;
+
+import java.util.stream.*;
+
+class EMailsTests {
+
+	@Property(tries = 10, edgeCases = EdgeCasesMode.NONE)
```
jlink, line 13: Since you run all properties with changed edgeCasesMode you can use `@ProperyDefaults(edgeCases=NONE)` on the class level.

engine/src/main/java/net/jqwik/engine/facades/ArbitrariesFacadeImpl.java
```diff
@@ -140,6 +140,59 @@ public StringArbitrary strings() {
 		return new DefaultStringArbitrary();
 	}
 
+	@Override
+	public Arbitrary<String> emails() {
```
jlink, line 145: I know we are going full circle, but now is probably a good time to introduce class `DefaultEmailArbitrary`. An interface for it is not needed yet IMO.

engine/src/test/java/net/jqwik/api/EMailsTests.java
jlink: 3. BTW, all the `replace` calls you make here should probably be `replaceAll` but that won't matter if go for something like `assertThat(domain.chars()).allMatch(c -> ALLOWED_DOMAIN_CHARS.contains(c));`

engine/src/test/java/net/jqwik/api/EMailsTests.java
```diff
@@ -0,0 +1,204 @@
...
+		Assertions.assertThat(InetAddresses.isInetAddress(domain)).isTrue();
+	}
+
+	@Property(edgeCases = EdgeCasesMode.NONE)
+	void quotedAndUnquotedUsernamesAreGenerated(@ForAll("emails") String email){
+		String localPart = getLocalPartOfEmail(email);
+		Statistics.collect(isQuoted(localPart));
+		Statistics.coverage(coverage -> {
+			coverage.check(true).count(c -> c >= 1);
+			coverage.check(false).count(c -> c >= 1);
```
jlink, line 145: Maybe percentage checks `.percentage(p -> p > 40%)` describe better what you want?

engine/src/test/java/net/jqwik/api/EMailsTests.java
zinki97: 3.: I think that .replace is also correct in this case

engine/src/main/java/net/jqwik/engine/facades/ArbitrariesFacadeImpl.java
zinki97: Is my basic implementation correct?

engine/src/test/java/net/jqwik/api/EMailsTests.java
jlink: ad 3. I probably don't understand the current code: How would multiple 'a's be removed?

zinki97: .replace("a", "") remove all "a"s

For example:
`System.out.println("aaaabcaaaadaaae".replace("a", ""));`
prints:
`bcde`

jlink: you're right. Had the impression that replace would have semantics of replaceFirst.
Anyway, you should get rid of that whole code all the same.

zinki97: Yes, this is my next step.
You're right, the replace methods are always confusing for me, too.

engine/src/test/java/net/jqwik/api/EMailsTests.java
```diff
@@ -29,54 +33,27 @@ void validLengthAfterAt(@ForAll("emails") String email){
...
+	}
+
+	@Property
+	void validUseOfQuotedBackslashAndQuotationMarks(@ForAll("emails") String email){
+		String localPart = getLocalPartOfEmail(email);
+		Assume.that(isQuoted(localPart));
+		localPart = localPart.substring(1, localPart.length() - 1);
+		localPart = localPart.replace("\\\"", "").replace("\\\\", "");
+		Assertions.assertThat(localPart).doesNotContain("\\");
+		Assertions.assertThat(localPart).doesNotContain("\"");
```
zinki97, line 56: Is this replace method okay? Because I don't know how I can check that localPart only contains \\" instead of " and \\\\ instead of \\ only with Assertions.

engine/src/test/java/net/jqwik/api/EMailsTests.java
```diff
@@ -216,6 +216,55 @@ void domainsWithOneAndMorePartsAreGenerated(@ForAll("emails") String email){
...
+		void domainShrinking(@ForAll Random random){
+			Arbitrary<String> emails = Arbitraries.emails().filter(v -> domainShrinkingFilter(v));
+			assertAllValuesAreShrunkTo("a@a", emails, random);
+		}
+
+		@Property
+		void domainShrinkingWithTLD(@ForAll Random random){
+			Arbitrary<String> emails = Arbitraries.emails().filter(v -> domainShrinkingWithTLDFilter(v));
+			assertAllValuesAreShrunkTo("a@a.aa", emails, random);
+		}
```
zinki97, line 232: The runtime of this is very high (32s), the runtime of domainShrinking is also high (8s).
I think the problem is the generation of ip addresses. Do you know what I can do?

jlink: Shrinking is a (comparatively) costly procedure. Especially with arbitraries with a lot of combinators, filters etc.

I suggest you reduce the number of tries for shrinking tests to 100 or even 10.

zinki97: The tries are not the problem as the tests are not working now (1 trie) and the runtime is so long

jlink: One problem is that it shrinks (correctly) to "A@a.aa" because "A" is smaller (by asci code)  that "a"

jlink: Another problem is 
```
private Arbitrary<String> domain(){
		return Arbitraries.frequencyOf(
				Tuple.of(1, domainIPv4()),
				Tuple.of(1, domainIPv6()),
				Tuple.of(2, domainDomain())
		);
	}
```
`Tuple.of(2, domainDomain())` should be the first option. Then it will be the preferred shrinking target.

zinki97: How can you change the shrinking order? And then reduce the runtime? Or should I change the testcases to "A@..."?

jlink: One more thing. The filtering should (in this case) be through a falsifier, e.g.

```
@Property
void ipv4Shrinking(@ForAll Random random) {
	Arbitrary<String> emails = Arbitraries.emails();
	Falsifier<String> falsifier = email -> {
		String domain = getDomainOfEmail(email);
		if (isIPAddress(domain) && domain.contains("."))
			return TryExecutionResult.falsified(null);
		else
			return TryExecutionResult.satisfied();
	};
	String value = shrinkToMinimal(emails, random, falsifier);
	assertThat(value).isEqualTo("A@[0.0.0.0]");
}
```

This will greatly improve perfromance of tests

engine/src/test/java/net/jqwik/api/EMailsTests.java
```diff
@@ -219,48 +220,70 @@ void domainsWithOneAndMorePartsAreGenerated(@ForAll("emails") String email){
...
-		@Property
+		@Property(tries = 20)
 		void domainShrinkingWithTLD(@ForAll Random random){
-			Arbitrary<String> emails = Arbitraries.emails().filter(v -> domainShrinkingWithTLDFilter(v));
-			assertAllValuesAreShrunkTo("a@a.aa", emails, random);
+			Arbitrary<String> emails = Arbitraries.emails();
+			Falsifier<String> falsifier = falsifyDomainWithTLD();
+			String value = shrinkToMinimal(emails, random, falsifier);
+			assertThat(value).isEqualTo("A@a.aa");
 		}
```
zinki97: I don't know why this is not working. Do you have an idea?
`Expecting:
     <"A@a.a.aa">
    to be equal to:
     <"A@a.aa">
    but was not.`

engine/src/test/java/net/jqwik/api/EMailsTests.java
```diff
@@ -219,48 +220,70 @@ void domainsWithOneAndMorePartsAreGenerated(@ForAll("emails") String email){
...
-		@Property
+		@Property(tries = 20)
 		void ipv6Shrinking(@ForAll Random random){
-			Arbitrary<String> emails = Arbitraries.emails().filter(v -> ipv6ShrinkingFilter(v));
-			assertAllValuesAreShrunkTo("a@[::0:0:0:0:0:0]", emails, random);
+			Arbitrary<String> emails = Arbitraries.emails();
+			Falsifier<String> falsifier = falsifyIPv6();
+			String value = shrinkToMinimal(emails, random, falsifier);
+			assertThat(value).isEqualTo("A@[::0:0:0:0:0:0]");
 		}
```
zinki97: How can I fix this problem?
`Expecting:
     <"A@[0:0:0:0:0::0:0]">
    to be equal to:
     <"A@[::0:0:0:0:0:0]">
    but was not.`

jlink: If "[::0:0:0:0:0:0]" is a valid ipv6 address then the shrunk result is correct. So is it?

engine/src/test/java/net/jqwik/api/EMailsTests.java
jlink: This shrinks to the correct value _most of the time_. Recursive shrinking is sometimes not fully stable because all shrinking depends on the initially created failing sample.

You could loosen the assertions a bit e.g. 
- Number of dots <= 3 (that's what I observed) and
- domain parts can only be "a" except for the TLD which must be "aa"

Getting a generator to _always_ shrink as expected is sometimes not possible and sometimes not worth it.

engine/src/test/java/net/jqwik/api/EMailsTests.java
zinki97: Yes, it is valid

engine/src/test/java/net/jqwik/api/EMailsTests.java
zinki97: Problem: Number of dots <= 3 does not work. I tried it with 10000 tries and got 4 dots. I think I only check the second point.

engine/src/test/java/net/jqwik/api/EMailsTests.java
jlink: Then check that all present parts are "0" and that there are 6 of them. Would that work?

zinki97: My version now consists in checking that each part is '0' or blank and that it is a valid IP address.

zinki97:
The runtime of `.withChars("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789-.")` is 4s faster than `.alpha().numeric().withChars("-.")`

Should I change it to `.alpha().numeric().withChars("-.")`?

jlink:
Your call.

Am Mo., 9. Nov. 2020 um 15:03 Uhr schrieb zinki97 <notifications@github.com

Issue #130, opened 2020-11-11
Add default generation for String with `@Email` annotation


jlink:
## Testing Problem

There's currently no way to generate emails just by an annotation.

## Suggested Solution

Register `EmailArbitraryProvider` that reacts to annotation `@Email`.

## Discussion

In the long run `@Email` should have some properties to configure types of email e.g.
`@Email(allowIpv6Domain = false)`

zinki97:
Working on annotation: https://github.com/jlink/jqwik/pull/131

jlink:
Merged PR. Available in 1.3.9-SNAPSHOT. 

See https://jqwik.net/docs/snapshot/user-guide.html#email-address-generation

PR #131, opened 2020-11-12
Email Annotation

zinki97:
See https://github.com/jlink/jqwik/issues/130

### DIFF
--- a/engine/src/main/java/net/jqwik/engine/providers/EmailArbitraryProvider.java
+++ b/engine/src/main/java/net/jqwik/engine/providers/EmailArbitraryProvider.java
@@ -10,14 +10,16 @@ public class EmailArbitraryProvider implements ArbitraryProvider {
 
 	@Override
 	public boolean canProvideFor(TypeUsage targetType) {
-		Optional<Email> annotation = targetType.findAnnotation(Email.class);
-		return !annotation.equals(Optional.empty());
+		return targetType.isAssignableFrom(String.class) && targetType.findAnnotation(Email.class).isPresent();
 	}
 
-	@SuppressWarnings({"rawtypes", "unchecked"})
 	@Override
 	public Set<Arbitrary<?>> provideFor(TypeUsage targetType, ArbitraryProvider.SubtypeProvider subtypeProvider) {
 		return Collections.singleton(Arbitraries.emails());
 	}
 
+	@Override
+	public int priority() {
+		return 5;
+	}
 }
--- a/engine/src/main/java/net/jqwik/engine/providers/EmailArbitraryProvider.java
+++ b/engine/src/main/java/net/jqwik/engine/providers/EmailArbitraryProvider.java
@@ -21,7 +21,7 @@ public class EmailArbitraryProvider implements ArbitraryProvider {
 		boolean allowDomains = email.allowDomains();
 		boolean allowIPv4 = email.allowIPv4();
 		boolean allowIPv6 = email.allowIPv6();
-		return Collections.singleton(Arbitraries.emails(allowQuotedLocalPart, allowUnquotedLocalPart, allowDomains, allowIPv4, allowIPv6));
+		return Collections.singleton(Arbitraries.emails());
 	}
 
 	@Override
--- a/engine/src/main/java/net/jqwik/engine/providers/EmailArbitraryProvider.java
+++ b/engine/src/main/java/net/jqwik/engine/providers/EmailArbitraryProvider.java
@@ -1,6 +1,7 @@
 package net.jqwik.engine.providers;
 
 import net.jqwik.api.*;
+import net.jqwik.api.arbitraries.*;
 import net.jqwik.api.constraints.*;
 import net.jqwik.api.providers.*;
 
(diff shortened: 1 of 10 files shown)


engine/src/main/java/net/jqwik/engine/providers/EmailArbitraryProvider.java
```diff
@@ -0,0 +1,23 @@
...
+
+import java.util.*;
+
+public class EmailArbitraryProvider implements ArbitraryProvider {
+
+	@Override
+	public boolean canProvideFor(TypeUsage targetType) {
+		Optional<Email> annotation = targetType.findAnnotation(Email.class);
+		return !annotation.equals(Optional.empty());
+	}
```
zinki97, line 15: I don't know if this is correct, because I have the following problem:

```
@Property(tries = 10)
void containsAtSign2(@ForAll String email) {
   System.out.println(email);
   //assertThat(email).contains("@");
}
```

Prints now strings like `ጆ衠⋓铨 䟹蘈쭱 禛` or `ꖇ≲`

When I add the Email annotation it prints sometimes correct addresses but sometimes wrong strings. Example:
```
"-EuEa#^A}"@[a04e:6:B:A:9:0:954:4d3c]
쯡᱾ 꺒뮭㜻衔ᴗ 崅ᣆ袽抜 咘周ɠ灀 ꤌꞿ鳘 ා븢 ∟끏豕 푟 腞☉쵊뢣㫾夬  탕娆⑩ﲏ 䚆ᱭ྄멈쭹✺ 쟤 㑸녴茲堡뼥 䄗 逴獷᧾ ⟙ሦ핦▉먲࡞䜘 筝霞㭏뉔꡻ 嶃  梗ȇ귕ƪᇰ
dW_-c&@[D:F:ED:F:C0c:ae::0]
帠ᮩ 較䘐אּ꬗㰓
```

jlink: The effect you see is that more than one arbitrary provider "declares itself" as suitable. In addition to yours it's `StringArbitraryProvider`. Therefore, jqwik chooses randomly among email and plain string generation.

What you have to do is to override method `int priority()` with a value > 0; I suggest 5.
Moreover, the canProvideFor-method should also check for type String. Similar to this:

```
@Override
public boolean canProvideFor(TypeUsage targetType) {
	return targetType.isAssignableFrom(String.class) && 
				   targetType.findAnnotation(Email.class).isPresent();
}

@Override
public int priority() {
    return 5;
}
```

zinki97: For `@ForAll @Email String email` it works now. But for `@ForAll String email` it prints again strings like
```
芺㳾 릔࠾ ࡊ 䇞
王ᩥ氁ᧂ숪  
釕앾
 宫 
儅䍹攓䄽
 뚄땎ꙺ 
ﺩ툴醨췝㢅敌뗧
෢ 痩닖⹲熌  鴣Ὴ
힕쑸䢲  氁䣜ω㎃ 
 竨聵ጰ洕  䋖䒈籪릲侓◦
```

jlink: I see no way how `@ForAll String email` could _reliably_ use the email arbitrary. The compilation name of a parameter is only available if you switch it on with a dedicated compiler switch (`-parameters`). That's why you cannot rely on that.

zinki97: I think I didn't explain it clearly enough. The problem is that `@ForAll String string` prints such a string. So the `@ForAll String` breaks and I don't know why.

zinki97: I've now seen version 1.3.8 print the same strings. Maybe that's not a problem at all? I thought it was a problem.

jlink: An unrestricted string can generate most unicode codepoints. Since there are many more Chinese ones than others those are the most probable to be generated. So this is expected behaviour.

zinki97: Okay, thanks :) Then I had a mistake in my thinking.

engine/src/main/java/net/jqwik/engine/providers/EmailArbitraryProvider.java
```diff
@@ -15,7 +15,13 @@ public boolean canProvideFor(TypeUsage targetType) {
 
 	@Override
 	public Set<Arbitrary<?>> provideFor(TypeUsage targetType, ArbitraryProvider.SubtypeProvider subtypeProvider) {
-		return Collections.singleton(Arbitraries.emails());
+		Email email = targetType.findAnnotation(Email.class).get();
```
zinki97, line 18: Is this okay to get the email annotation? It works in the testcases.

jlink: Thats correct this way.

jlink (changes requested):
My main point is that Arbitraries.emails() should return object of type EmailArbitrary. Otherwise users of the programmatic API cannot use all the advanced features like quotedLocalParts() etc.

This two ways to use the API should also be reflected in the tests. Whereas ArbitrariesEmailsTests should (mostly) use the programmatic API (Arbitraries.emails()...) there should be an additional test container class in package net.jqwik.api.constraints (EmailProperties) that just checks that the corresponding features of EmailArbitrary can be invoked through the Email annotation.

zinki97:
I updated it. Is it correct now?

jlink:
Looks almost perfect now. I'm missing one property in `EmailsProperties` for default behaviour of `@Email`.

Apart from that the PR is ready for merging.

zinki97:
> Looks almost perfect now. I'm missing one property in `EmailsProperties` for default behaviour of `@Email`.
> 
> Apart from that the PR is ready for merging.

Should it check the statistics or what should it check?

jlink:
> > Looks almost perfect now. I'm missing one property in `EmailsProperties` for default behaviour of `@Email`.
> > Apart from that the PR is ready for merging.
> 
> Should it check the statistics or what should it check?

Sounds like a good idea.

zinki97:
> Looks almost perfect now. I'm missing one property in `EmailsProperties` for default behaviour of `@Email`.
> 
> Apart from that the PR is ready for merging.

Done

jlink (approved):

PR #135, opened 2020-11-23
Not all valid IPv6 addresses can be generated

zinki97:
Addresses like a::a can't be generated

### DIFF
--- a/engine/src/test/java/net/jqwik/api/ArbitrariesEmailsTests.java
+++ b/engine/src/test/java/net/jqwik/api/ArbitrariesEmailsTests.java
@@ -88,6 +88,7 @@ public class ArbitrariesEmailsTests {
 			Assume.that(!isIPAddress(domain));
 			assertThat(domain.charAt(0)).isNotEqualTo('-');
 			assertThat(domain.charAt(domain.length() - 1)).isNotEqualTo('-');
+			assertThat(domain).contains(".");
 			assertThat(domain).doesNotContain("..");
 			assertThat(domain.charAt(0)).isNotEqualTo('.');
 			assertThat(domain.charAt(domain.length() - 1)).isNotEqualTo('.');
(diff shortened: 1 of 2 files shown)


zinki97:
Ready for review

engine/src/test/java/net/jqwik/api/ArbitrariesEmailsTests.java
```diff
@@ -116,7 +115,7 @@ void tldNotAllNumeric(@ForAll("emails") String email) {
 			assertThat(stringContainsMinimumOneChar(ALLOWED_NOT_NUMERIC_CHARS_TLD, tld)).isTrue();
 		}
 
-		@Property
+		@Property(tries = 5000)
```
jlink, line 118: Why would you try that 5000 times?

jlink (approved):

engine/src/test/java/net/jqwik/api/ArbitrariesEmailsTests.java
zinki97: To have a greater chance that addresses like a::a are generated

engine/src/test/java/net/jqwik/api/ArbitrariesEmailsTests.java
```diff
@@ -313,11 +316,20 @@ void ipv6Shrinking(@ForAll Random random) {
 			String value = shrinkToMinimal(emails, random, falsifier);
 			String domain = getDomainOfEmail(value);
 			domain = domain.substring(1, domain.length() - 1);
-			String[] domainParts = domain.split(":");
-			assertThat(isValidIPv6Address(domain)).isTrue();
-			IntStream.range(0, domainParts.length).forEach(i -> {
-				assertThat(domainParts[i]).isIn("0", "");
-			});
+			assertThat(domain).isEqualTo("::");
```
jlink, line 319: parts of ipv6 should always shrink to "0", should they not?

PR #136, opened 2020-11-25
Fix generate single sample with null value

mhyeon-lee:
## Overview

`Arbitrary.sample` throws NullPointException when generating null values.
Calling the `findFirst` method on Stream does not allow null values.

### Details

To avoid handling null values, wrap it with Optional and handle it.

---

### DIFF
--- a/api/src/main/java/net/jqwik/api/Arbitrary.java
+++ b/api/src/main/java/net/jqwik/api/Arbitrary.java
@@ -441,8 +441,10 @@ public interface Arbitrary<T> {
 	@API(status = MAINTAINED, since = "1.3.0")
 	default T sample() {
 		return this.sampleStream()
-				   .findFirst()
-				   .orElseThrow(() -> new JqwikException("Cannot generate a value"));
+			.map(Optional::ofNullable)
+			.findFirst()
+			.orElseThrow(() -> new JqwikException("Cannot generate a value"))
+			.orElse(null);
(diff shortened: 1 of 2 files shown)


jlink:
@mhyeon-lee Thanks for the catch. 

I cannot merge the PR, though, unless you add 

```
---

```

to the buttom of the PRs description.

Thank you.

mhyeon-lee:
@jlink I added agreement description. thank you

jlink:
@mhyeon-lee Many thanks!

api/src/main/java/net/jqwik/api/Arbitrary.java
```diff
@@ -441,8 +441,10 @@ public boolean isUnique() {
 	@API(status = MAINTAINED, since = "1.3.0")
 	default T sample() {
 		return this.sampleStream()
-				   .findFirst()
-				   .orElseThrow(() -> new JqwikException("Cannot generate a value"));
+			.map(Optional::ofNullable)
```
luvarqpp, line 444: I would be a little bit concerned about performance... On the other side, this method should not be called extensively. It is called only by user while experimenting, or perhaps from "normal" JUnit tests. Am I right?

jlink:
To be frank I'm not concerned. When generating values so much is going on that the little bit of overhead through mapping should be neglectible. But I didn't measure.

Issue #140, opened 2020-12-09
Add jqwik Module for Generation of Dates and Times


jlink:
## Testing Problem

Generating dates and times with certain constraints is a standard problem when building domain-specific
generators. Currently jqwik requires users to generate dates from numbers or other primitive data types.

## Suggested Solution

Provide a new module/artefact (id "jqwik-time") that has means to generate time  and date values.
The new module will be maintained together with jqwik core and might be a default dependency when including
the aggregate artefact "jqwik".

Foundation for generation are the types in java.time. Mappers to "old" JDK classes like Date, Timestamp, Calendar etc. should be provided for convenience.

Base package for the module is `net.jqwik.api.time`. There are two classes that serve as containers for factory methods:
- `net.jqwik.api.time.Times` with method `TimeArbitrary times()`
- `net.jqwik.api.time.Dates` with method `DateArbitrary dates()`

In the long run default generation by type should be possible for all date types and time types, including `Duration` and `Period`. There should also be annotations available for the most common time and date constraints, e.g. before, after, betwee etc.

### Date Generation Options

We ignore the beginning of the Gregorian Calendar (Feb 24 1582)

- Between start/end date (borders in- or excluded)
- Min/Max Year
- Min/Max Month
- Min/Max Day of month (what's the use-case?)
- Constrain day of week
- Generate just a year
- Generate just a month
- Generate just a day of week
- Generate a month in a year

#### Optional

- With/without lap year/lap day
- Generate Period (date-based time span)
- Map to Date
- Map to Calendar

### Time Generation Options

- Min/Max Hours
- Min/Max Minutes
- Min/Max Seconds
- Min/Max Millis
- Between given times (borders in- or excluded)
- Constrain precision (nanos, millis, seconds, minutes)
- Include/exclude lap seconds

### Instant, LocalDateTime, OffsetDateTime, ZonedDateTime

Combine date and time constraints with (optional) offset/timezone/dst generation.
Handle impossible times, e.g. between 2 and 3 am on change to DST.

## Annotations

To be discussed

## Discussion

What about non Gregorian dates?

jlink:
dsfgsdfg

zinki97:
Which class should be used for the date? I think the Date class is deprecated. Maybe Calendar?

Calendar and Date also have time attributes. How should we handle this?

zinki97:
My first design of DateArbitrary:
```
public interface DateArbitrary extends Arbitrary<LocalDate> {

[...]

PR #141, opened 2020-12-10
Generation of Dates

zinki97:
See https://github.com/jlink/jqwik/issues/140


time/src/main/java/net/jqwik/time/DefaultDateArbitrary.java
```diff
@@ -70,6 +71,13 @@ public DateArbitrary monthLessOrEqual(Month max) {
 		return clone;
 	}
 
+	@Override
+	public DateArbitrary onlyMonths(Month... months) {
+		DefaultDateArbitrary clone = typedClone();
+		clone.allowedMonths = months;
+		return clone;
+	}
```
zinki97, line 79: How should I handle the value null?

time/src/test/java/net/jqwik/api/time/MonthTests.java
```diff
@@ -0,0 +1,112 @@
...
+
+	@Group
+	class CheckMonthMethods {
+
+		@Property
+		void atTheEarliest(@ForAll("months") Month month) {
+
+			Arbitrary<Month> months = Dates.months().atTheEarliest(month);
+
+			assertAllGenerated(months.generator(1000), m -> {
```
jlink, line 33: Use net.jqwik.api.testing.TestingSupport.assertAllGenerated(...)

Get a random throuch `@ForAll Random random`.

time/src/main/java/net/jqwik/time/DefaultDateArbitrary.java
jlink: You mean if months is null?

That's not legal. But I don't think this is checked elsewhere.

time/src/main/java/net/jqwik/time/DefaultDateArbitrary.java
```diff
@@ -8,9 +8,147 @@
...
 	@Override
 	protected Arbitrary<LocalDate> arbitrary() {
-		return Arbitraries.just(LocalDate.of(2020, 12, 9));
+		optimizeRuntime();
+		Arbitrary<Year> year = generateYears();
+		Arbitrary<Month> month = generateMonths();
+		Arbitrary<Integer> dayOfMonth = generateDayOfMonths();
+		return Combinators.combine(year, month, dayOfMonth)
+						  .as(this::generateValidDateFromValues)
+						  .filter(v -> v != null && !v.isBefore(dateMin) && !v.isAfter(dateMax) && isInAllowedDayOfWeeks(v.getDayOfWeek()));
```
jlink, line 30: Heavy filtering is usually bad for exhaustive generation, edge case generation and shrinking. I wonder if tuning year, month and dayOfMonth generation could get rid of some of the filter parts.

time/src/main/java/net/jqwik/time/DefaultMonthDayArbitrary.java
```diff
@@ -0,0 +1,80 @@
...
+	protected Arbitrary<MonthDay> arbitrary() {
+		DateArbitrary dates = Dates.dates()
+								   .atTheEarliest(LocalDate.of(0, monthDayMin.getMonth(), monthDayMin.getDayOfMonth()))
+								   .atTheLatest(LocalDate.of(0, monthDayMax.getMonth(), monthDayMax.getDayOfMonth()))
+								   .yearBetween(0, 0)
+								   .monthBetween(monthMin, monthMax)
+								   .onlyMonths(allowedMonths)
+								   .dayOfMonthBetween(dayOfMonthMin, dayOfMonthMax);
+		return dates.map(v -> MonthDay.of(v.getMonth(), v.getDayOfMonth()));
+	}
```
jlink, line 30: this looks very complicated for "only" creating a valid date within a year.
Can year 0 be a lap year? Should Feb 29 be possible.

time/src/test/java/net/jqwik/api/time/DaysOfMonthTests.java
```diff
@@ -0,0 +1,81 @@
...
+import java.util.*;
+
+import net.jqwik.api.*;
+
+import static org.assertj.core.api.Assertions.*;
+
+import static net.jqwik.api.testing.TestingSupport.*;
+
+@Group
+public class DaysOfMonthTests {
```
jlink, line 12: test classes need not be public

jlink (changes requested):
Some general remarks:
- Please adapt your local code style to jqwik standard and reformat.
- Please add tests for shrinking
- Please add tests for edge cases
- Please add tests for exhaustive generation

time/src/main/java/net/jqwik/time/DefaultMonthDayArbitrary.java
zinki97: Yes, year 0 can be a lap year. Should I change it? I thought it is good to use existing code instead of writing a completely new arbitrary.

time/src/main/java/net/jqwik/api/time/DateArbitrary.java
```diff
@@ -5,4 +5,51 @@
 import net.jqwik.api.*;
 
 public interface DateArbitrary extends Arbitrary<LocalDate> {
+
+	//TODO: Documentation
```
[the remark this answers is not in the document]
zinki97: yeah I know, I'll add it later

time/src/main/java/net/jqwik/time/DefaultDateArbitrary.java
zinki97: The `optimizeRuntime` method makes it possible to filter as few dates as possible.

I put the other methods in `generateDateFromValues`. Is it better now?

time/src/main/java/net/jqwik/time/DefaultMonthDayArbitrary.java
jlink: > Yes, year 0 can be a lap year. Should I change it? I thought it is good to use existing code instead of writing a completely new arbitrary.

Reuse is good... if it doesn't complicate things. Generating dates and stripping away the year will lead to bad performance concerning shrinking and exhaustive genration and lead to duplicate edge cases.

time/src/main/java/net/jqwik/time/DefaultDateArbitrary.java
```diff
@@ -8,9 +8,147 @@
...
+
+	private Arbitrary<Month> generateMonths(){
+		return Dates.months().between(monthMin, monthMax).only(allowedMonths);
+	}
+
+	private Arbitrary<Integer> generateDayOfMonths(){
+		return Dates.daysOfMonth().between(dayOfMonthMin, dayOfMonthMax);
+	}
+
+	private LocalDate generateValidDateFromValues(Year y, Month m, int d){
```
jlink: Can probably be replaced with a call to `Arbitrary.ignoreException(..)`.

time/src/main/java/net/jqwik/time/DefaultMonthDayArbitrary.java
zinki97: even if the year is always 0?

time/src/main/java/net/jqwik/time/DefaultDateArbitrary.java
jlink: Maybe `optimizeRuntime` is just a name I don't understand. It looks to me like some of the things therein should either be done when setting the values or be checked when setting the values for inconsistency - like minYear being greater than maxYear.

time/src/main/java/net/jqwik/api/time/DaysOfWeekArbitrary.java
```diff
@@ -0,0 +1,12 @@
+package net.jqwik.api.time;
+
+import java.time.*;
+
+import net.jqwik.api.*;
+
+public interface DaysOfWeekArbitrary extends Arbitrary<DayOfWeek> {
```
jlink, line 7: Since DayOfWeek is just an enum the whole arbitrary class is not necessary or is it. Could be replaced with Arbitraries.of(DayOfWeek.class) I think. only(..) doe not pull its weight either since Arbitraries.of(MONDAY, WEDNESDAY) is just as short.

time/src/main/java/net/jqwik/time/DefaultMonthDayArbitrary.java
jlink: Not sure. Have you done an experiment to check?

time/src/main/java/net/jqwik/api/time/DaysOfWeekArbitrary.java
zinki97: Okay, I'll remove it

time/src/main/java/net/jqwik/time/DefaultMonthDayArbitrary.java
zinki97: No, but `Dates.years().between(yearMin, yearMax)` creates only one year wit the value 0. Only `generateMonths` and `generateDayOfMonths` have more values. So the year value is always the same and can't change.

zinki97:
Can you please provide feedback on the `ExhaustiveGeneration` classes?

time/src/test/java/net/jqwik/api/time/DatesTests.java
```diff
@@ -1,20 +1,342 @@
...
+			assertThat(value).isEqualTo(LocalDate.of(0, Month.JANUARY, 1));
+		}
+
+		@Property
+		void shrinksToSmallestFailingPositiveValue(@ForAll Random random){
+			DateArbitrary dates = Dates.dates();
+			TestingFalsifier<LocalDate> falsifier = date -> date.isBefore(LocalDate.of(2013, Month.MAY, 25));
+			LocalDate value = shrinkToMinimal(dates, random, falsifier);
+			assertThat(value).isEqualTo(LocalDate.of(2013, Month.MAY, 25));
+		}
```
zinki97, line 278: How can I fix this?

`Expecting:
     <2014-01-01 (java.time.LocalDate)>
    to be equal to:
     <2013-05-25 (java.time.LocalDate)>
    but was not.`

Or should I accept that it can't shrink perfectly?

jlink: I'd aim for perfect shrinking here. To see why it won't happen requires detailed analysis. Leave it as is, and create a list with things that don't work as expected yet.

jlink:
> Can you please provide feedback on the `ExhaustiveGeneration` classes?

Tests for exhaustive generation look fine to me. 

I'd probably also look at a full generation for all MonthAndDay values. But I may have missed that.

time/src/test/java/net/jqwik/api/time/DatesTests.java
zinki97: List is here: https://github.com/jlink/jqwik/issues/140#issuecomment-743322513

time/src/test/java/net/jqwik/api/time/DatesTests.java
```diff
@@ -1,20 +1,384 @@
...
+																																																															 .of(2020, Month.DECEMBER, 21), LocalDate
+																																																																									.of(2020, Month.DECEMBER, 24), LocalDate
+																																																																																		   .of(2020, Month.DECEMBER, 28), LocalDate
+																																																																																												  .of(2020, Month.DECEMBER, 31));
+		}
+
+	}
+
+	@Group
+	class EdgeCasesTests {
```
zinki97, line 343: Can you please provide feedback on the `EdgeCasesTests` classes?

Some of the tests are currently not working.
For example:
```
@Property(tries = 5)
void between() {

	MonthArbitrary months = Dates.months().between(Month.MARCH, Month.AUGUST);
	Set<Month> edgeCases = collectEdgeCases(months.edgeCases());
	assertThat(edgeCases).hasSize(2);
	assertThat(edgeCases).contains(Month.MARCH, Month.AUGUST);

}
```

jlink: What's the error?

zinki97:
```
java.lang.AssertionError:
    Expected size:<2> but was:<0> in:
    <[]>
```

jlink: I'll have a closer look. 
BTW, `@Property(tries = 5)` doesnt make sense IMO since edge case generation is deterministic.
I'd go with `@Example`

jlink: If you take away the "smallest" and "largest" month, no edge case remains which is what I'd expect. So I think you're good here.

jlink: Similiar to edge cases, exhaustive generation is also deterministic and thus I'd go with `@Example` there.

time/src/test/java/net/jqwik/api/time/MonthTests.java
```diff
@@ -0,0 +1,204 @@
...
+			MonthArbitrary months = Dates.months().between(Month.MARCH, Month.AUGUST);
+			Set<Month> edgeCases = collectEdgeCases(months.edgeCases());
+			assertThat(edgeCases).hasSize(2);
+			assertThat(edgeCases).contains(Month.MARCH, Month.AUGUST);
+
+		}
+
+	}
+
+	public static Arbitrary<Month[]> generateMonths() {
```
jlink, line 181: Maybe I dont undestand it, but if it's supposed to generate an Array of Month, I'd do it with `monthArbitrary.array(Month[].class).ofSize...`

time/src/test/java/net/jqwik/api/time/DatesTests.java
zinki97: `months = months.edgeCases(monthConfig -> monthConfig.add(min).add(max));`

I fixed it with that and I am working on the other edge cases. Should I remove it?

zinki97: As user I expect the min and max values as edgecases

jlink: Using `EdgeCases.Config.add(..)` has the drawback that values added this way cannot be shrunk. 
Given there are only 12 months anyway, I'd drop it here.

jlink: > As user I expect the min and max values as edgecases

You could manipulate the original selection to go from min to max. Then those edge cases would get in automatically.

zinki97: > Using `EdgeCases.Config.add(..)` has the drawback that values added this way cannot be shrunk.
> Given there are only 12 months anyway, I'd drop it here.

Sure that this is a problem? The shrinking tests are not bader now.

zinki97: I'll upload it so you can take a closer look.

time/src/test/java/net/jqwik/api/time/MonthTests.java
zinki97: The arbitrary generates arrays with different months and lengths

jlink: So array() is the right thing I guess.

time/src/main/java/net/jqwik/time/DefaultMonthArbitrary.java
```diff
@@ -0,0 +1,57 @@
...
+import net.jqwik.api.time.*;
+
+public class DefaultMonthArbitrary extends ArbitraryDecorator<Month> implements MonthArbitrary {
+
+	private Month min = Month.JANUARY;
+	private Month max = Month.DECEMBER;
+	private Month[] allowedMonths = new Month[]{Month.JANUARY, Month.FEBRUARY, Month.MARCH, Month.APRIL, Month.MAY, Month.JUNE, Month.JULY, Month.AUGUST, Month.SEPTEMBER, Month.OCTOBER, Month.NOVEMBER, Month.DECEMBER};
+
+	@Override
+	protected Arbitrary<Month> arbitrary() {
```
jlink, line 16: I suggest the following implementation

	@Override
	protected Arbitrary<Month> arbitrary() {
		List<Month> values = Arrays.asList(Month.class.getEnumConstants());
		final int indexMin = values.indexOf(min);
		final int indexMax = values.indexOf(max);

		return Arbitraries.integers()
						  .between(indexMin, indexMax)
						  .edgeCases(integerConfig -> integerConfig.includeOnly(indexMin, indexMax))
						  .map(values::get)
						  .filter(this::isInAllowedMonths);
	}

This will break one exhaustive generation test because this generation is more precise.

time/src/test/java/net/jqwik/api/time/MonthDayTests.java
```diff
@@ -0,0 +1,333 @@
...
+
+			}
+
+			@Property
+			void monthOnlyMonths(@ForAll("monthsOnlyMonths") Month[] months, @ForAll Random random) {
+
+				Arbitrary<MonthDay> dates = Dates.monthDays().onlyMonths(months);
+
+				assertAllGenerated(dates.generator(1000), random, md -> {
+					assertThat(md.getMonth()).isIn(months);
```
jlink, line 144: This generates a vararg warning. Use List<Month> instead of Month[]. This happens in a few other places as well.

time/src/test/java/net/jqwik/api/time/MonthTests.java
jlink: This is my suggestion:

	public static Arbitrary<Month[]> generateMonths() {
		Arbitrary<Month> monthArbitrary = Arbitraries.of(Month.class);
		return monthArbitrary.array(Month[].class).ofMinSize(1);
	}

Better even, use a `Set<Month>` but then you don't need the whole method anymore becaus you can just say ` @ForAll @Size(min=1) Set<Month>` at point of usage.

time/src/main/java/net/jqwik/time/DefaultMonthArbitrary.java
jlink: It can even be further simplified if we drop `MonthArbitrary.includeOnly(..)` which is just a complicated way of saying `Arbitrary.of(..)`.

time/src/test/java/net/jqwik/api/time/MonthDayTests.java
zinki97: I think the warning has now disappeared by using Set <month>.

time/src/main/java/net/jqwik/time/DefaultDateArbitrary.java
```diff
@@ -8,9 +8,135 @@
 
 public class DefaultDateArbitrary extends ArbitraryDecorator<LocalDate> implements DateArbitrary {
```
jlink, line 9: All Default* classes should get a
`@API(status = INTERNAL)` annotation.

time/src/main/java/net/jqwik/api/time/DateArbitrary.java
```diff
@@ -2,7 +2,72 @@
...
+
+	/**
+	 * Set the allowed lower {@code min} (included) and upper {@code max} (included) bounder of generated year values.
+	 */
+	DateArbitrary yearBetween(Year min, Year max);
+
+	/**
+	 * Set the allowed lower {@code min} (included) and upper {@code max} (included) bounder of generated year values.
+	 */
+	default DateArbitrary yearBetween(int min, int max) {
```
jlink, line 42: Min/max int values should be documented.

time/src/main/java/net/jqwik/api/time/DateArbitrary.java
```diff
@@ -2,7 +2,72 @@
...
+
+	/**
+	 * Set the allowed lower {@code min} (included) and upper {@code max} (included) bounder of generated month values.
+	 */
+	DateArbitrary monthBetween(Month min, Month max);
+
+	/**
+	 * Set the allowed lower {@code min} (included) and upper {@code max} (included) bounder of generated month values.
+	 */
+	default DateArbitrary monthBetween(int min, int max) {
```
jlink, line 54: Min/max int values should be documented.

time/src/main/java/net/jqwik/api/time/DateArbitrary.java
```diff
@@ -2,7 +2,72 @@
...
+
+	/**
+	 * Set an array of allowed {@code months}.
+	 */
+	DateArbitrary onlyMonths(Month... months);
+
+	/**
+	 * Set the allowed lower {@code min} (included) and upper {@code max} (included) bounder of generated day of month values.
+	 */
+	DateArbitrary dayOfMonthBetween(int min, int max);
```
jlink, line 66: Min/max int values should be documented.

time/src/main/java/net/jqwik/api/time/DaysOfMonthArbitrary.java
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * Fluent interface to configure the generation of day of month values.
+ */
+@API(status = EXPERIMENTAL, since = "1.4.0")
+public interface DaysOfMonthArbitrary extends Arbitrary<Integer> {
+
+	/**
+	 * Set the allowed lower {@code min} (included) and upper {@code max} (included) bounder of generated day of month values.
+	 */
+	default DaysOfMonthArbitrary between(int min, int max) {
```
jlink, line 18: Min/max int values should be documented.

time/src/main/java/net/jqwik/api/time/DaysOfMonthArbitrary.java
```diff
@@ -0,0 +1,32 @@
...
+		return greaterOrEqual(min).lessOrEqual(max);
+	}
+
+	/**
+	 * Set the allowed lower {@code min} (included) bounder of generated day of month values.
+	 */
+	DaysOfMonthArbitrary greaterOrEqual(int min);
```
jlink, line 25: Min/max int values should be documented.

time/src/main/java/net/jqwik/api/time/DaysOfMonthArbitrary.java
```diff
@@ -0,0 +1,32 @@
...
+
+	/**
+	 * Set the allowed upper {@code max} (included) bounder of generated day of month values.
+	 */
+	DaysOfMonthArbitrary lessOrEqual(int max);
```
jlink, line 30: Min/max int values should be documented.

time/src/main/java/net/jqwik/api/time/MonthArbitrary.java
```diff
@@ -0,0 +1,60 @@
...
+	 * Set the allowed lower {@code min} (included) and upper {@code max} (included) bounder of generated month values.
+	 */
+	default MonthArbitrary between(Month min, Month max) {
+		return atTheEarliest(min).atTheLatest(max);
+	}
+
+	/**
+	 * Set the allowed lower {@code min} (included) and upper {@code max} (included) bounder of generated month values.
+	 */
+	default MonthArbitrary between(int min, int max) {
```
jlink, line 27: Min/max int values should be documented.

time/src/main/java/net/jqwik/api/time/MonthArbitrary.java
```diff
@@ -0,0 +1,60 @@
...
+
+	/**
+	 * Set the allowed upper {@code max} (included) bounder of generated month values.
+	 */
+	MonthArbitrary atTheLatest(Month max);
+
+	/**
+	 * Set the allowed upper {@code max} (included) bounder of generated month values.
+	 */
+	default MonthArbitrary atTheLatest(int max) {
```
jlink, line 51: Min/max int values should be documented.

time/src/main/java/net/jqwik/api/time/MonthDayArbitrary.java
```diff
@@ -0,0 +1,94 @@
...
+	 * Set the allowed lower {@code min} (included) and upper {@code max} (included) bounder of generated month values.
+	 */
+	default MonthDayArbitrary monthBetween(Month min, Month max) {
+		return monthGreaterOrEqual(min).monthLessOrEqual(max);
+	}
+
+	/**
+	 * Set the allowed lower {@code min} (included) and upper {@code max} (included) bounder of generated month values.
+	 */
+	default MonthDayArbitrary monthBetween(int min, int max) {
```
jlink, line 44: Min/max int values should be documented.

time/src/main/java/net/jqwik/api/time/MonthDayArbitrary.java
```diff
@@ -0,0 +1,94 @@
...
+
+	/**
+	 * Set the allowed lower {@code min} (included) bounder of generated month values.
+	 */
+	MonthDayArbitrary monthGreaterOrEqual(Month min);
+
+	/**
+	 * Set the allowed lower {@code min} (included) bounder of generated month values.
+	 */
+	default MonthDayArbitrary monthGreaterOrEqual(int min) {
```
jlink, line 56: Min/max int values should be documented.

time/src/main/java/net/jqwik/api/time/MonthDayArbitrary.java
```diff
@@ -0,0 +1,94 @@
...
+	 * Set the allowed lower {@code min} (included) and upper {@code max} (included) bounder of generated day of month values.
+	 */
+	default MonthDayArbitrary dayOfMonthBetween(int min, int max) {
+		return dayOfMonthGreaterOrEqual(min).dayOfMonthLessOrEqual(max);
+	}
+
+	/**
+	 * Set the allowed lower {@code min} (included) bounder of generated day of month values.
+	 */
+	MonthDayArbitrary dayOfMonthGreaterOrEqual(int min);
```
jlink, line 87: Min/max int values should be documented.

time/src/main/java/net/jqwik/api/time/MonthDayArbitrary.java
```diff
@@ -0,0 +1,94 @@
...
+
+	/**
+	 * Set the allowed upper {@code max} (included) bounder of generated day of month values.
+	 */
+	MonthDayArbitrary dayOfMonthLessOrEqual(int max);
```
jlink, line 92: Min/max int values should be documented.

time/src/main/java/net/jqwik/api/time/MonthDayArbitrary.java
```diff
@@ -0,0 +1,94 @@
...
+
+import net.jqwik.api.*;
+
+import static org.apiguardian.api.API.Status.*;
+
+/**
+ * Fluent interface to configure the generation of month and day values.
+ */
+@API(status = EXPERIMENTAL, since = "1.4.0")
+public interface MonthDayArbitrary extends Arbitrary<MonthDay> {
```
jlink, line 15: I'd get rid of all the greate/less/orEqual methods. Too much API maintenance burden for too little value.

time/src/main/java/net/jqwik/api/time/YearArbitrary.java
```diff
@@ -0,0 +1,55 @@
...
+
+import net.jqwik.api.*;
+
+import static org.apiguardian.api.API.Status.*;
+
+/**
+ * Fluent interface to configure the generation of Year values.
+ */
+@API(status = EXPERIMENTAL, since = "1.4.0")
+public interface YearArbitrary extends Arbitrary<Year> {
```
jlink, line 15: I'd get rid of all the greate/less/orEqual methods. Too much API maintenance burden for too little value.

time/src/main/java/net/jqwik/api/time/YearMonthArbitrary.java
```diff
@@ -0,0 +1,115 @@
...
+
+import net.jqwik.api.*;
+
+import static org.apiguardian.api.API.Status.*;
+
+/**
+ * Fluent interface to configure the generation of year and month values.
+ */
+@API(status = EXPERIMENTAL, since = "1.4.0")
+public interface YearMonthArbitrary extends Arbitrary<YearMonth> {
```
jlink, line 15: I'd get rid of all the greate/less/orEqual methods. Too much API maintenance burden for too little value.

documentation/src/docs/user-guide.template.md
```diff
@@ -1295,6 +1295,55 @@ void restrictedEmailAddresses(@ForAll @Email(quotedLocalPart = false, ipv4Host =
 }
 ```
 
+### Date Arbitrary Types
```
jlink, line 1298: Looking at the other headlines I'd rather write "Generation of Dates".

Using annotations (as soons as there are any) will also be documented here so a more generic headline fits better.

documentation/src/docs/user-guide.template.md
```diff
@@ -1295,6 +1295,55 @@ void restrictedEmailAddresses(@ForAll @Email(quotedLocalPart = false, ipv4Host =
...
+
```
jlink, line 1299: Add a sentence to describe that date generation is in an extra module which must be in a project's dependency.

documentation/src/docs/user-guide.template.md
```diff
@@ -1295,6 +1295,55 @@ void restrictedEmailAddresses(@ForAll @Email(quotedLocalPart = false, ipv4Host =
...
+- [`YearArbitrary years()`](/docs/${docsVersion}/javadoc/net/jqwik/api/Dates.html#years())
+- [`MonthArbitrary months()`](/docs/${docsVersion}/javadoc/net/jqwik/api/Dates.html#months())
+- [`DaysOfMonthArbitrary daysOfMonth()`](/docs/${docsVersion}/javadoc/net/jqwik/api/Dates.html#daysOfMonth())
+- [`YearMonthArbitrary yearMonths()`](/docs/${docsVersion}/javadoc/net/jqwik/api/Dates.html#yearMonths())
+- [`MonthDayArbitrary monthDays()`](/docs/${docsVersion}/javadoc/net/jqwik/api/Dates.html#monthDays())
+
+In addition, you can constrain their values using the following functions:
+
+#### DateArbitrary
+
```
jlink, line 1312: Add a code example or a few for generation of dates. Mention that `LocalDate` is the target type.

jlink (changes requested):

jlink (changes requested):
1. Some test classes no longer compile. I suggest to run clean Gradle build before pushing changes.

2. There are no tests checking statistical distribution. I think they would make sense to check that months, day of week, and day of months are approximately equally distributed.

3. Javadoc of DateArbitrary should mention that all generated dates use the Gregorian Calendar, even if they are before Oct 15, 1582.

4. I'd like to have default min year to be 1900 and max year 2500. So by default only dates between Jan 1, 1900 and Dec 31, 2500 are generated. This must be documented. That change is otional; I can add it myself later.

zinki97:
- [ ] 1.
- [ ] 2.
- [ ] 3.
- [ ] 4.
- [ ] 5.

Issue #139, opened 2020-12-09
Replace jqwik.properties with JUnit Platform Mechanism for Configuration


jlink:
## Testing Problem

jqwik has its own way to configure itself: `jqwik.properties`. This has a couple of disadvantages:
- Users must be taught how to use it
- Users must create such a file in each of their projects
- There's currently no mechanism to override configuration parameters, e.g. through system properties or runtime parameters

## Suggested Solution

Switch to the [configuration mechanism](https://junit.org/junit5/docs/current/user-guide/#running-tests-config-params) built into the JUnit platform. 

For versions 1.4.x the old mechanism must be kept for backwards compatibility.

The first implementation step should cover:

- Configuration properties can be loaded from `jqwik.properties` or from `junit-platform.properties` with an additional "jqwik." prefix
- Loading from `jqwik.properties` will log a WARNING
- Having a parameter in both file will log an ERROR. Parameter from `junit-platform.properties` has precedence
- Rename some of the config parameters for `junit-platform.properties` (log an ERROR if old name is used there)
   - defaultTries -> jqwik.tries.default
   - defaultMaxDiscardRatio -> jqwik.maxDiscardRatio.default
   - defaultGeneration -> jqwik.generation.default
   - defaultAfterFailure -> jqwik.afterFailure.default
   - defaultEdgeCases -> jqwik.edgeCases.default
   - defaultShrinking -> jqwik.shrinking.default
- Optional: Provide a tool to translate `jqwik.properties` into `junit-platform.properties`

## Discussion

- Currently jqwik gets its configuration when creating the instance of `JqwikTestEngine`. 
  The `ConfigurationParameters` instance is only available in discovery and execution requests. 
  Maybe a first step should move the retrieval of `JqwikConfiguration` into the request methods to
  facilitate migration.

- JUnit's `ConfigurationParameters` interface does not provide a means to get all parameters with a certain prefix.
That means that it will no longer be possible to warn about unused/unknown parameters (e.g. through spelling errors).

- Should camel case names be replaced by names with hyphens or underscores?

osi:
On property naming, the JUnit Jupiter documentation does dot-separated lowercase. they use `testinstance` and `autodetection` instead of a camel-cased or adding other separators. a dot appears to be used for grouping (as it does with the proposed `.default` syntax).

osi:
[...]

PR #145, opened 2020-12-22
load properties using junit configuration parameters mechanism

osi:
## Overview

Implement #139. Properties are now loaded from JUnit's `ConfigurationParameters`. Backwards compatibility rules are implemented per spec. 

### Details

- When using the IntelliJ code format w/2020.3, it wanted to change the continuation indent in some places. 
- No additional tests were added as a part of this (as this portion of the code didn't have existing tests, other than for default values). Tests didn't seem to be high-value here. 
- I updated the doc template; It wasn't clear if I needed to do something to regenerate the HTML documentation.
- Unsure if there should be a stated end to the compatibility property loading.

---


engine/src/main/java/net/jqwik/engine/JqwikTestEngine.java
```diff
@@ -82,4 +75,12 @@ private void registerLifecycleHooks(TestDescriptor rootDescriptor, Configuration
 		new JqwikLifecycleRegistrator(lifecycleRegistry, configurationParameters).registerLifecycleHooks(rootDescriptor);
 	}
 
+	JqwikConfiguration buildConfiguration() {
```
jlink, line 78: Not private because?

jlink (commented):
Looks good so far. 

Do you prefer to continue working in this PR or should I merge in smaller steps?

engine/src/test/java/net/jqwik/engine/TestDescriptorBuilder.java
```diff
@@ -91,7 +91,7 @@ public TestDescriptor build(TestDescriptor parent) {
 
 	private TestDescriptor createDescriptor(TestDescriptor parent) {
 		if (element instanceof JqwikTestEngine)
-			return new JqwikEngineDescriptor(engineId());
+			return new JqwikEngineDescriptor(engineId(), ((JqwikTestEngine)element).buildConfiguration());
```
osi, line 94: @jlink re: https://github.com/jlink/jqwik/pull/145/files#r548860845

`buildConfiguration` wasn't private because I was calling it here (in order to propagate the configuration). I admittedly didn't dig into how jqwik's tests use itself to run. I've modified this to pass in `null`, as all the tests continue to pass.

osi:
> Do you prefer to continue working in this PR or should I merge in smaller steps?

No strong preference. I'll be building on top of this when continuing the work. If you'd rather wait for the entirety of the parameter work to be complete before merging, no objections.

jlink:
Currently it seems easier to continue this branch and PR. If at some point I'd like to touch code that's also being touched here, we can change policy.

osi:
Sounds good. I'll push works-in-progress as I have incremental points towards completion.

engine/src/test/resources/junit-platform.properties
```diff
@@ -0,0 +1,11 @@
...
+#jqwik.runfailuresfirst=false
+#jqwik.tries.default=1000
+#jqwik.maxdiscardratio.default=5
+#jqwik.usejunitplatformreporter=false
+#jqwik.generation.default=AUTO
+#jqwik.afterfailure.default=PREVIOUS_SEED
+jqwik.reportonlyfailures=true
+jqwik.edgecases.default=FIRST
+#jqwik.shrinking.default=OFF
+jqwik.boundedshrinkingseconds=20
```
jlink, line 11: Maybe `jqwik.shrinking.bounded.seconds` as key.

engine/src/test/resources/junit-platform.properties
```diff
@@ -0,0 +1,11 @@
+#jqwik.database=.jqwik-database
+#jqwik.runfailuresfirst=false
+#jqwik.tries.default=1000
+#jqwik.maxdiscardratio.default=5
+#jqwik.usejunitplatformreporter=false
```
jlink, line 5: `jqwik.reporting.usejunitplatform`

engine/src/test/resources/junit-platform.properties
```diff
@@ -0,0 +1,11 @@
...
+#jqwik.generation.default=AUTO
+#jqwik.afterfailure.default=PREVIOUS_SEED
+jqwik.reportonlyfailures=true
```
jlink, line 8: `jqwik.reporting.onlyfailures`

jlink (commented):
I have the feeling that properties keys could do with a bit more structure. Consider my remarks to be suggestions. Maybe you have better ones.

osi:
> I have the feeling that properties keys could do with a bit more structure. Consider my remarks to be suggestions. Maybe you have better ones.

No problem, it does make sense to use this as an opportunity to add more structure. 

I also updated the encapsulation of the compatibility logic.

jlink:
@osi Anything else you plan to do within this PR?

osi:
@jlink No, I believe that this is complete.

Issue #138, opened 2020-12-04
Ability to ignore (or fail) on usage of @Property(seed=)


osi:
Being able to add the `seed` argument to the `@Property` annotation is superb for reproducing failures.

However, sometimes they get accidentally committed, which can lead to a CI server runnings tests on a fixed seed. 

The ability to ignore or fail on the presence of an explicit seed, for use in CI systems, would help prevent this error. 

A system property or environment variable feels to be the easiest to make conditional upon the environment (as users may have `jqwik.properties` committed to SCM)

jlink:
There’s two aspects here:
1. Configuring special behaviour when seed is specified (use, warn, fail, ignore)
2. Overriding config values in system prop 

1 is crucial for you. 2 is optional, right?

osi:
Correct. My goal is to prevent builds running in certain environments from using a fixed seed. (developer laptop, ok. CI server, no)

jlink:
I’m a bit in doubt if a global (overwritable) config is all that helpful.  I’ve seen a few codebases where some properties are purposefully fixed to a seed for a reason. Those properties should IMO not lead to failures. 

What about having a second attribute in @Property, e.g. temporarySeed, which will always log a warning? I’m just rambling. Maybe/probably there are better options to reconcile both scenarios.

osi:
What if the source of the seed for a property was available in a `AroundPropertyHook` ? It could be as simple of a change as having `PropertyExecutionResult#seed` return a tuple of the seed and the source (perhaps an enum of `PREVIOUS_SEED`, `RANDOM_SEED`, `FIXED_SEED`). A hook could transition the result to failed if it was using a `FIXED_SEED` in an inappropriate execution environment.

jlink:
@osi Extending the AroundPropertyHook with such a specific trait does not feel adequate.

Here's my current suggestion:
1. Add a configuration parameter to `jqwik.properties`:
   `onFixedSeed: ALLOW | WARN | FAIL`
2. Allow configuration parameters be overridden by system properties handed in via
   `-D jqwik.onFixedSeed=FAIL`
3. Extend configuration to allow something like:
   `onFixedSeed.net.myproject.MyTestclass.myPropertyMethod=ALLOW`

What do you think?

Maybe jqwik should switch the whole configuration approach to use junit-platform.properties as described 
[here](https://junit.org/junit5/docs/current/user-guide/#running-tests-config-params). 
[...]

PR #149, opened 2021-01-20
add ability to warn or fail a test if a seed is fixed

osi:
introduce a `FixedSeedMode` that can be set globally (via properties)
or on an individual test. this allows control of using the `seed`
argument on a property, as it can be undesirable to accidentally commit
a value in this field after reproducing a failure.

fix for #138

### DIFF
--- a/engine/src/main/java/net/jqwik/engine/execution/CheckedProperty.java
+++ b/engine/src/main/java/net/jqwik/engine/execution/CheckedProperty.java
@@ -59,25 +59,24 @@ public class CheckedProperty {
 		try {
 			effectiveConfiguration = configurationWithEffectiveSeed();
 		} catch (FailOnFixedSeedException failOnFixedSeedException) {
-			return PropertyCheckResult.failed(
-					configuration.getStereotype(), propertyName, 0, 0,
-					configuration.getSeed(), configuration.getGenerationMode(),
-					configuration.getEdgeCasesMode(), 0, 0,
-					null, null, failOnFixedSeedException
-			);
+			return failed(configuration, failOnFixedSeedException);
 		}
 		maybeWarnOnMultipleTriesWithoutForallParameters(effectiveConfiguration);
 		try {
 			Reporter reporter = propertyLifecycleContext.reporter();
 			return createGenericProperty(effectiveConfiguration).check(reporter, reporting);
 		} catch (CannotFindArbitraryException cannotFindArbitraryException) {
-			return PropertyCheckResult.failed(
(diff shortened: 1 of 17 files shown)


engine/src/main/java/net/jqwik/engine/execution/CheckedProperty.java
```diff
@@ -55,7 +55,17 @@ public CheckedProperty(
...
+		try {
+			effectiveConfiguration = configurationWithEffectiveSeed();
+		} catch (FailOnFixedSeedException failOnFixedSeedException) {
+			return PropertyCheckResult.failed(
+					configuration.getStereotype(), propertyName, 0, 0,
+					configuration.getSeed(), configuration.getGenerationMode(),
+					configuration.getEdgeCasesMode(), 0, 0,
+					null, null, failOnFixedSeedException
+			);
+		}
```
osi, line 68: This catch block feels funny. I can't merge it with the next though due to it needing the actual `effectiveConfiguration`

jlink: Cannot see a more elegant solution either atm. Sometimes bad looking code has to stay until s.o. suddenly has an epiphany.

jlink:
@osi Do you think it's ready for merging?

osi:
I believe it's ready. The CI failure seems like a permission issue of some sort.

jlink:
Yes, CI is on me. 

Von meinem iPad gesendet

jlink (changes requested):
If you don't find the time I can do the few changes myself.

osi:
> If you don't find the time I can do the few changes myself.

I'll take care of them in the next few days. I appreciate the suggestions!

engine/src/main/java/net/jqwik/engine/execution/CheckedProperty.java
```diff
@@ -55,7 +55,17 @@ public CheckedProperty(
 	}
 
 	public PropertyCheckResult check(Reporting[] reporting) {
-		PropertyConfiguration effectiveConfiguration = configurationWithEffectiveSeed();
+		PropertyConfiguration effectiveConfiguration;
+		try {
```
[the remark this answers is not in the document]
osi: The problem with moving into that method is that the try-catch block translates the exception to a `PropertyCheckResult`. 

I extracted the handling of the exception into a helper method, as the same work is done twice. perhaps that's sufficient for now?

jlink:
Many thanks. Going to merge it.

Issue #154, opened 2021-01-26
Add Generation of Times


zinki97:
## Testing Problem

Generating times with certain constraints is a standard problem when building domain-specific
generators. Currently jqwik requires users to generate times from numbers or other primitive data type.

## Suggested Solution

Add arbitraries to the new module/artefact (id "jqwik-time") to generate time values.

Foundation for generation are the types in java.time. Mappers to "old" JDK classes like Date, Timestamp, Calendar etc. should be provided for convenience.

Add a new class to the base package net.jqwik.api.time:

- net.jqwik.api.time.Times

In the long run default generation by type should be possible for all time types, including Duration. There should also be annotations available for the most common time constraints, e.g. before, after, between etc.

zinki97:
### Time Generation Options
* [ ] Min/Max Hours
* [ ] Min/Max Minutes
* [ ] Min/Max Seconds
* [ ] Min/Max Milliseconds
* [ ] Min/Max Nanoseconds
* [ ] Between given times (borders in- or excluded)
* [ ] Constrain precision (nanos, millis, seconds, minutes)
* [ ] Include/exclude lap seconds

### Default Generation of Dates
* [ ] LocalTime
* [ ] Duration

### Annotations for Time
* [ ] `@TimeRange`
* [ ] `@HourRange`
* [ ] `@MinuteRange`
* [ ] `@SecondRange`
* [ ] `@MillisecondRange`
* [ ] `@NanosecondRange`

zinki97:
Design idea:

class `Times`:
- `LocalTimeArbitrary times()`
- `DurationArbitrary durations()`

class `LocalTimeArbitrary`:
TODO

jlink:
I would provide more options in class `Times`:
- `LocalTimeArbitrary localTimes()`
- `OffsetTimeArbitrary offsetTimes()`
- `LocalDateTimeArbitrary localDateTimes()`
- `DateArbitrary localDateTimesAsDate()`
- `InstantArbitrary instants()`
- `OffsetDateTimeArbitrary offsetDateTimes()`
- `ZonedDateTimeArbitrary zonedDateTimes()`

And probably one of:
- `ZoneIdArbitrary zoneIds()`
- `TimeZoneArbitrary timeZones()`
I'm not sure which is more generic and useful.

zinki97:
I think it's better to add DateTimes in a separate step and in this step only Times.

jlink:
[...]

PR #156, opened 2021-01-27
Generation of Times

zinki97:
See https://github.com/jlink/jqwik/issues/154


zinki97:
I've tried a similar solution to `DefaultLocalDateArbitrary`, but I think it is too much filtering here.

I try a new solution.

time/src/test/java/net/jqwik/time/api/LocalTimeTests.java
```diff
@@ -20,12 +20,28 @@
...
+
+		@Property
+		void worstCaseTimeGeneration(@ForAll("worstCases") LocalTime time) {
+			assertThat(time).isNotNull();
+		}
+
+		@Provide
+		Arbitrary<LocalTime> worstCases() {
+			return Times.times().between(LocalTime.of(22, 59, 59, 999_999_998), LocalTime.of(23, 00, 00, 000_000_001));
+		}
```
zinki97, line 39: Is it a problem when this worst case scenario does not work?

```
timestamp = 2021-01-27T17:24:36.556964900, SimpleArbitraries:worstCaseTimeGeneration = 
  net.jqwik.api.TooManyFilterMissesException:
    Filtering [Filtering [net.jqwik.api.RandomGenerator$$Lambda$403/0x0000000800c6c440@36ebc363]] missed more than 10000 times.
```

zinki97: I think I have an idea: 2 algorithms that generate

jlink: Between() shouldn’t require filtering IMO so the concrete example failing may not be a problem but the general approach is.

zinki97:
I need more testcases. I'll add them later.

time/src/test/java/net/jqwik/time/api/LocalTimeTests.java
zinki97: The problem is using plusNanos causes a lot of filtering. So I can't use the approach from `DefaultLocalDateArbitrary`.

jlink: A single generator of just nanos will probably not work. One idea is to have to generators, one for seconds within a day and another one for nanos within a second. Very often the second one will not be used when only second accuracy is required.

time/src/main/java/net/jqwik/time/internal/properties/arbitraries/DefaultLocalTimeArbitrary.java
```diff
@@ -19,72 +20,42 @@
...
+			effectiveMin = calculateEffectiveMin();
+		} catch (NoValuesCanBeGeneratedException e) {
+			//TODO: Other Exception??
+			throw new TooManyFilterMissesException(e.getMessage());
+		}
 		LocalTime effectiveMax = calculateEffectiveMax();
+		if (effectiveMax.isBefore(effectiveMin)) {
+			//TODO: Other Exception??
+			throw new TooManyFilterMissesException("The maximum time is too soon after the minimum time.");
+		}
```
zinki97, line 46: Is this a good exception or should I throw another exception?

time/src/test/java/net/jqwik/time/api/LocalTimeTests.java
```diff
@@ -352,6 +482,7 @@ void defaultShrinking(@ForAll Random random) {
 			assertThat(value).isEqualTo(LocalTime.of(0, 0, 0, 0));
 		}
 
+		//TODO: Too much time?
 		@Property
 		void shrinksToSmallestFailingValue(@ForAll Random random) {
```
zinki97, line 487: This test case works, but need 44 seconds on my machine. What should I do? Constrain the precision to Millis/Micros/Seconds?

jlink: Reducing precision is an option. I also wonder whatthe default precision should be. I tend to seconds mit maybe millis is more what someone would expect?

zinki97: That's an idea. At the moment default precision is Nanos

time/src/test/java/net/jqwik/time/api/LocalTimeTests.java
```diff
@@ -482,13 +483,12 @@ void defaultShrinking(@ForAll Random random) {
...
 
-		//TODO: Too much time?
 		@Property
 		void shrinksToSmallestFailingValue(@ForAll Random random) {
-			LocalTimeArbitrary times = Times.times();
+			LocalTimeArbitrary times = Times.times().constrainPrecision(SECONDS);
 			TestingFalsifier<LocalTime> falsifier = time -> time.isBefore(LocalTime.of(9, 13, 42, 143_921_111));
 			LocalTime value = falsifyThenShrink(times, random, falsifier);
-			assertThat(value).isEqualTo(LocalTime.of(9, 13, 42, 143_921_111));
+			assertThat(value).isEqualTo(LocalTime.of(9, 13, 43, 0));
```
zinki97: If the testcase works it still needs 6 seconds. 
If it fails it needs only 60 milliseconds. Why is that?
Example:
```
Shrinking:shrinksToSmallestFailingValue = 
  org.opentest4j.AssertionFailedError:
    Expecting:
     <09:13:43>
    to be equal to:
     <09:13:42>
    but was not.
```

zinki97:
I don't know why the checks are failing. Have you an idea @jlink ?

jlink (commented):
> I don't know why the checks are failing. Have you an idea @jlink ?

I cannot reproduce the compile errors locally. All that I can currently think of is that it could be JDK-specific. Do you run Linux and can install 1.8.0_282 (Azul Systems, Inc. 25.282-b08)?

zinki97:
> > I don't know why the checks are failing. Have you an idea @jlink ?
> 
> I cannot reproduce the compile errors locally. All that I can currently think of is that it could be JDK-specific. Do you run Linux and can install 1.8.0_282 (Azul Systems, Inc. 25.282-b08)?

I use Windows

jlink:
> > > I don't know why the checks are failing. Have you an idea @jlink ?
> > 
> > 
> > I cannot reproduce the compile errors locally. All that I can currently think of is that it could be JDK-specific. Do you run Linux and can install 1.8.0_282 (Azul Systems, Inc. 25.282-b08)?
> 
> I use Windows

I have an idea what's happening here. Github takes the test support module dependency from some cache. Since the method does no longer exist (it was renamed) it cannot resolve it. This is either a bug in Github Actions or a misconfiguration on my side. Suggested solution: Rebase your branch on the current main and fix the breaking code. I think this should fix.

BTW, it's a good idea to rebase from current main regularly - first thing in the morning - to prevent your code moving too far away from main.

time/src/main/java/net/jqwik/time/internal/properties/arbitraries/DefaultLocalTimeArbitrary.java
zinki97: How can I say that no values can be generated with this configuration at the runtime instead?

jlink: In the end I guess an IllegalArgumentException is the right one, cause this shows that arbitrary has been wrongly configured, right?

time/src/main/java/net/jqwik/time/internal/properties/arbitraries/DefaultLocalTimeArbitrary.java
```diff
@@ -0,0 +1,309 @@
...
+
+	private ChronoUnit precision = MILLIS;
+
+	@Override
+	protected Arbitrary<LocalTime> arbitrary() {
+
+		LocalTime effectiveMin;
+		try {
+			effectiveMin = calculateEffectiveMin();
+		} catch (NoValuesCanBeGeneratedException e) {
```
jlink, line 38: What's the point of NoValuesCanBeGeneratedException? maybe you can use an IllegalArgumentException at the origin of the throws?

zinki97:
Most of the tests are failing now, even in `EmailsTests` and `LocalDateTests`

Example Errors:
LocalTimeTests:
```
Minutes:precisionMaxTimeSoonAfterMinTime = 
  java.lang.AbstractMethodError:
    Receiver class net.jqwik.engine.facades.ArbitraryFacadeImpl$2 does not define or inherit an implementation of the resolved method 'abstract net.jqwik.api.EdgeCases edgeCases(int)' of interface net.jqwik.api.Arbitrary.
```

EmailTests:
```
AllGeneratedEmailAddressesAreValid:validCharsAfterAt = 
  java.lang.NoSuchMethodError:
    'net.jqwik.api.RandomGenerator net.jqwik.api.RandomGenerator.flatMap(java.util.function.Function, int)'
```

time/src/main/java/net/jqwik/time/internal/properties/arbitraries/DefaultLocalTimeArbitrary.java
zinki97: Is an IllegalArgumentException okay at the runtime?

jlink: It's made for that purpose and being thrown throughout jqwik.

jlink:
> Most of the tests are failing now, even in `EmailsTests` and `LocalDateTests`
> 
> Example Errors:
> LocalTimeTests:
> 
> ```
> Minutes:precisionMaxTimeSoonAfterMinTime = 
>   java.lang.AbstractMethodError:
>     Receiver class net.jqwik.engine.facades.ArbitraryFacadeImpl$2 does not define or inherit an implementation of the resolved method 'abstract net.jqwik.api.EdgeCases edgeCases(int)' of interface net.jqwik.api.Arbitrary.
> ```
> 
> EmailTests:
> 
> ```
> AllGeneratedEmailAddressesAreValid:validCharsAfterAt = 
>   java.lang.NoSuchMethodError:
>     'net.jqwik.api.RandomGenerator net.jqwik.api.RandomGenerator.flatMap(java.util.function.Function, int)'
> ```

What you checked in doesn't even compile here. I suggest a full and clean rebuild.

zinki97:
> > Most of the tests are failing now, even in `EmailsTests` and `LocalDateTests`
> > Example Errors:
> > LocalTimeTests:
> > ```
> > Minutes:precisionMaxTimeSoonAfterMinTime = 
> >   java.lang.AbstractMethodError:
> >     Receiver class net.jqwik.engine.facades.ArbitraryFacadeImpl$2 does not define or inherit an implementation of the resolved method 'abstract net.jqwik.api.EdgeCases edgeCases(int)' of interface net.jqwik.api.Arbitrary.
> > ```
> > 
> > 
> > EmailTests:
> > ```
> > AllGeneratedEmailAddressesAreValid:validCharsAfterAt = 
> >   java.lang.NoSuchMethodError:
> >     'net.jqwik.api.RandomGenerator net.jqwik.api.RandomGenerator.flatMap(java.util.function.Function, int)'
> > ```
> 
> What you checked in doesn't even compile here. I suggest a full and clean rebuild.

The checked in version is not my current version. I'll push my current version.

jlink:
> The checked in version is not my current version. I'll push my current version.

I see 12 failing tests now but none seems to be due to missing method or abstract method error.

zinki97:
> > > Most of the tests are failing now, even in `EmailsTests` and `LocalDateTests`
> > > Example Errors:
> > > LocalTimeTests:
> > > ```
> > > Minutes:precisionMaxTimeSoonAfterMinTime = 
> > >   java.lang.AbstractMethodError:
> > >     Receiver class net.jqwik.engine.facades.ArbitraryFacadeImpl$2 does not define or inherit an implementation of the resolved method 'abstract net.jqwik.api.EdgeCases edgeCases(int)' of interface net.jqwik.api.Arbitrary.
> > > ```
> > > 
> > > 
> > > EmailTests:
> > > ```
> > > AllGeneratedEmailAddressesAreValid:validCharsAfterAt = 
> > >   java.lang.NoSuchMethodError:
> > >     'net.jqwik.api.RandomGenerator net.jqwik.api.RandomGenerator.flatMap(java.util.function.Function, int)'
> > > ```
> > 
> > 
> > What you checked in doesn't even compile here. I suggest a full and clean rebuild.
> 
> The checked in version is not my current version. I'll push my current version.

Okay, it seems like the error is on my machine. But I've rebuilded it 15 minutes ago. Maybe I need to restart.

zinki97:
restart and then rebuild worked!

jlink:
> restart and then rebuild worked!

IntelliJ can be a b****

zinki97:
After many hours of work ZoneOffset generation with restrictions works :) I'll add some tests for invalid configurations and equal distribution, etc.

time/src/main/java/net/jqwik/time/internal/properties/arbitraries/DefaultOffsetTimeArbitrary.java
```diff
@@ -59,28 +62,50 @@
 			localTimes = localTimes.constrainPrecision(ofPrecision);
 		}
 
-		return localTimes.map(v -> OffsetTime.of(v, ZoneOffset.MIN));
+		if (offsetMin != null && offsetMax != null) {
+			zoneOffsets = zoneOffsets.between(offsetMin, offsetMax);
+		}
+
+		Arbitrary<OffsetTime> offsetTimes = Combinators.combine(localTimes, zoneOffsets).as(OffsetTime::of);
```
zinki97, line 69: Is it okay to use `Times.zoneOffsets().sample()` instead of `Combinators.combine` to combine value?

The problem is too heavy filtering and I don't know if it is possible to calculate good min/max values. Maybe with Times.zoneOffsets().sample() in the right range it is possible that there is not so much filtering... (See testcases)

jlink: I currently cannot see the code in its context but never use sample() within an arbitrary. You lose all shrinking capabilities and may run into strange effects since sample() is made for use outside properties.

zinki97: Should I accept then that bad values runs into TooManyFilterMissesException? Or have you an idea what I can do?

Tomorrow I'll experiment with LocalTime min/max value maybe it's better then.

time/src/main/java/net/jqwik/time/internal/properties/arbitraries/DefaultOffsetTimeArbitrary.java
```diff
@@ -0,0 +1,205 @@
...
+import static java.time.temporal.ChronoUnit.*;
+import static org.apiguardian.api.API.Status.*;
+
+@API(status = INTERNAL)
+public class DefaultOffsetTimeArbitrary extends ArbitraryDecorator<OffsetTime> implements OffsetTimeArbitrary {
+
+	private OffsetTime offsetTimeMin = OffsetTime.MIN;
+	private OffsetTime offsetTimeMax = OffsetTime.MAX;
+
+	private LocalTime timeMin = null;
```
jlink, line 22: Why don't you just delegate all the constraint values timeMinMax etc. to an embedded LocalTimeArbitrary?

time/src/main/java/net/jqwik/time/internal/properties/arbitraries/DefaultOffsetTimeArbitrary.java
```diff
@@ -0,0 +1,205 @@
...
+			localTimes = localTimes.constrainPrecision(ofPrecision);
+		}
+
+		if (offsetMin != null && offsetMax != null) {
+			zoneOffsets = zoneOffsets.between(offsetMin, offsetMax);
+		}
+
+		Arbitrary<OffsetTime> offsetTimes = Combinators.combine(localTimes, zoneOffsets).as(OffsetTime::of);
+
+		offsetTimes = offsetTimes.filter(v -> !(v.isBefore(offsetTimeMin) || v.isAfter(offsetTimeMax)));
```
jlink, line 71: Maybe you can translate offsetTimeMin/Max into the offset for an embedded localTimeArbitrary. This should get rid of all the fitering. 

Maybe it does not really make sense at all to constrain by OffsetTime instances, since one could choose min and max to have different ZoneIds. I see no point in that. If you choose to offer min and max of LocalTime (which implicitly use the generated ZoneId) all those difficulties do not arise.

time/src/main/java/net/jqwik/time/internal/properties/arbitraries/DefaultZoneOffsetArbitrary.java
```diff
@@ -0,0 +1,112 @@
...
+	public static final ZoneOffset DEFAULT_MIN = ZoneOffset.ofHoursMinutesSeconds(-12, 0, 0);
+	public static final ZoneOffset DEFAULT_MAX = ZoneOffset.ofHoursMinutesSeconds(14, 0, 0);
+
+	private ZoneOffset offsetMin = DEFAULT_MIN;
+	private ZoneOffset offsetMax = DEFAULT_MAX;
+
+	@Override
+	protected Arbitrary<ZoneOffset> arbitrary() {
+
+		ZoneOffset[] zoneOffsets = generateAllValues();
```
jlink, line 26: I suggest you just generate the number of 15-minute-intervals, positive and negative, and calculate the real offset in a final map operation.

time/src/main/java/net/jqwik/time/internal/properties/arbitraries/DefaultOffsetTimeArbitrary.java
zinki97: So you mean remove offsetTimeMin and offsetTimeMax? Or what do you mean?

jlink: Use borders of type LocalTime instead.

time/src/test/java/net/jqwik/time/api/ZoneOffsetTests.java
```diff
@@ -113,6 +113,41 @@ void betweenSame(@ForAll("offsets") ZoneOffset sameOffset, @ForAll Random random
...
+
+				Arbitrary<ZoneOffset> offsets = Times.zoneOffsets().between(startOffset, endOffset);
+
+				assertAllGenerated(offsets.generator(1000), random, offset -> {
+					assertThat(offset.getTotalSeconds()).isGreaterThanOrEqualTo(startOffset.getTotalSeconds());
+					assertThat(offset.getTotalSeconds()).isLessThanOrEqualTo(endOffset.getTotalSeconds());
+					return true;
+				});
+
+			}
```
zinki97, line 144: This testcase looks very complicated, but it is definitely neccessary. It founds two bugs!

jlink: :-) what do startB and endB stand for?

zinki97: I forgot to rename it. It says whether the offset is negative or positive.

time/src/main/java/net/jqwik/time/internal/properties/arbitraries/DefaultOffsetTimeArbitrary.java
zinki97: This helps! Thanks for the feedback :)

zinki97:
Could you please provide feedback to `OffsetTimeArbitrary`?

time/src/main/java/net/jqwik/time/internal/properties/arbitraries/DefaultOffsetTimeArbitrary.java
```diff
@@ -0,0 +1,180 @@
...
+import net.jqwik.time.api.*;
+import net.jqwik.time.api.arbitraries.*;
+
+import static java.time.temporal.ChronoUnit.*;
+import static org.apiguardian.api.API.Status.*;
+
+@API(status = INTERNAL)
+public class DefaultOffsetTimeArbitrary extends ArbitraryDecorator<OffsetTime> implements OffsetTimeArbitrary {
+
+	private LocalTime timeMin = LocalTime.MIN;
```
jlink, line 19: I still think that delegating all configuration to an embedded localTimeArbitrary and zoneIdArbitrary could save a lot of code.

jlink:
BTW, the new target version has changed to 1.5.1 since I’ll soon have to publish a 1.5.0 due to urgent compatibility breaking changes.

time/src/main/java/net/jqwik/time/internal/properties/arbitraries/DefaultOffsetTimeArbitrary.java
zinki97: How can I embed them? What exactly do you mean? Do you have an example?

jlink: The simplest example is probably DefaultBigIntegerArbitrary. The idea is that you delegate the configuration, e.g. between(min, max), to the embeded arbitrary:
```java
OffsetTimeArbitrary clone = this.cloneTyped();
clone.localTimeArbitrary = localTimeArbitrary.between(min, max);
return clone;
```

zinki97: Done.

time/src/test/java/net/jqwik/time/api/DurationTests.java
```diff
@@ -0,0 +1,503 @@
...
+		}
+
+		@Property
+		@Disabled("Failing at the moment.")
+		void shrinksToSmallestFailingPositiveValue(@ForAll Random random) {
+			DurationArbitrary durations = Times.durations();
+			TestingFalsifier<Duration> falsifier = duration -> duration.compareTo(Duration.ofSeconds(999_392_192, 709_938_291)) < 0;
+			Duration value = falsifyThenShrink(durations, random, falsifier);
+			assertThat(value).isEqualTo(Duration.ofSeconds(999_392_192, 709_938_291));
+		}
```
zinki97, line 282: Same old problem...: Should I accept that shrinking does not work perfectly or what should I do?

```
org.opentest4j.AssertionFailedError:
    expected: 277608H56M32.709938291S
    but was : 277608H56M33S
```

jlink: I think shrinking to nanos accuracy is not necessary. So it's fine with me.

zinki97: Now I have this error:
```
org.opentest4j.AssertionFailedError:
    expected: 277608H56M33S
    but was : 277608H56M32.709938291S
```

Should I check that the shrunk value is one of this: `277608H56M33S`, `277608H56M32.709938291S` and accept both values?

jlink: maybe <= *M33

zinki97: Both test cases works now, but `shrinksToSmallestFailingPositiveValue` needs `1m20s` and `shrinksToSmallestFailingNegativeValue` needs `2m51s`.

Should I use only 5-100 tries?

jlink: I think shrinking test cases are often run with just 100 tries

time/src/test/java/net/jqwik/time/api/DurationTests.java
```diff
@@ -496,7 +496,24 @@ void between6() {
...
+			Statistics.label("Seconds")
+					  .collect(duration.getSeconds())
+					  .coverage(this::checkSecondsCoverage);
+		}
+
+		private void checkSecondsCoverage(StatisticsCoverage coverage) {
+			for (int value = -10; value <= 10; value++) {
+				coverage.check(value).percentage(p -> p >= 2.0);
+			}
+		}
```
zinki97, line 511: Do you have any idea why this test case is failing?

```
[CheckEqualDistribution:seconds] (1000) Seconds = 
    -10 (79) :  8 %
    10  (79) :  8 %
    0   (70) :  7 %
    -1  (53) :  5 %
    -2  (52) :  5 %
    4   (51) :  5 %
    -7  (49) :  5 %
    -4  (47) :  5 %
    -3  (45) :  5 %
    7   (45) :  5 %
    -8  (45) :  5 %
    8   (45) :  5 %
    2   (44) :  4 %
    1   (43) :  4 %
    -5  (42) :  4 %
    9   (42) :  4 %
    3   (35) :  4 %
    5   (34) :  3 %
    -9  (34) :  3 %
    -6  (33) :  3 %
    6   (33) :  3 %
```

```
CheckEqualDistribution:seconds = 
  org.opentest4j.AssertionFailedError:
    Percentage of 0.0 for [-10] does not fulfill condition for label "Seconds"
```

jlink: Is checksSecondsCoverage() used in more than one place? Try if inlining the method will change the result.

zinki97: I've checked this already, it doesn't change the result:
```
Statistics.label("Seconds")
					  .collect(duration.getSeconds())
					  .coverage(c -> c.check(-5).percentage(p -> p >= 2.0));
```
`-5  (30) :  3 %`
`Percentage of 0.0 for [-5] does not fulfill condition for label "Seconds"`

zinki97: After some experimentation, I can say the following:
It has something to do with `duration.getSeconds()`

Fails:
```
Statistics.label("Seconds")
					  .collect(duration.getSeconds())
					  .coverage(c -> c.check(0).percentage(p -> p >= 2.0));
```
`0   (82) :  8 %`
`Percentage of 0.0 for [0] does not fulfill condition for label "Seconds"`

Works:
```
Statistics.label("Seconds")
					  .collect(duration.getNano())
					  .coverage(c -> c.check(0).percentage(p -> p >= 2.0));
```
`0         (57) :  5.70 %`

zinki97: I found the problem! int vs. long 

Feedback: Maybe jqwik should show that there is a comparison between int and long or show the datatype

jlink: How should jqwik know for which value you are looking? You may collect longs and ints in the same statistics. I'll see if comparison can be more lenient.

zinki97: Okay that's true. But maybe `-10L (79) :  8 %` or `-10 (79 long) :  8 %` instead of `-10 (79) :  8 %`

zinki97:
Could you please provide general feedback to all Arbitraries. What is currently missing: Annotations.

Otherwise I think Time Generation (without DateTimes) is finished.

jlink:
Over the weekend.

Am Fr., 19. Feb. 2021 um 16:47 Uhr schrieb zinki97 <notifications@github.com

time/src/main/java/net/jqwik/time/api/Times.java
```diff
@@ -0,0 +1,70 @@
...
+	public static ZoneOffsetArbitrary zoneOffsets() {
+		return new DefaultZoneOffsetArbitrary();
+	}
+
+	/**
+	 * Create an arbitrary that generates instances of {@linkplain java.util.TimeZone}.
+	 *
+	 * @return a new arbitrary instance
+	 */
+	public static Arbitrary<TimeZone> timeZones() {
```
jlink, line 48: It could be derived from a ZoneId, couldn't it?

time/src/main/java/net/jqwik/time/api/arbitraries/LocalTimeArbitrary.java
```diff
@@ -0,0 +1,61 @@
...
+	/**
+	 * Set the allowed lower {@code min} (included) and upper {@code max} (included) bounder of generated second values.
+	 * The minutes can be between {@code 0} and {@code 59}.
+	 */
+	LocalTimeArbitrary secondBetween(int min, int max);
+
+	/**
+	 * Constrain the precision of generated values.
+	 */
+	LocalTimeArbitrary constrainPrecision(ChronoUnit ofPrecision);
```
jlink, line 59: constrainPrecision -> ofPrecision 
Document default precision.

time/src/main/java/net/jqwik/time/api/arbitraries/OffsetTimeArbitrary.java
```diff
@@ -0,0 +1,67 @@
...
+	/**
+	 * Set the allowed lower {@code min} (included) and upper {@code max} (included) bounder of generated offset values.
+	 * The offsets can be between {@code -12:00:00} and {@code +14:00:00}.
+	 */
+	OffsetTimeArbitrary offsetBetween(ZoneOffset min, ZoneOffset max);
+
+	/**
+	 * Constrain the precision of generated values.
+	 */
+	OffsetTimeArbitrary constrainPrecision(ChronoUnit ofPrecision);
```
jlink, line 65: ofPrecision(..)
document default

time/src/main/java/net/jqwik/time/api/arbitraries/ZoneOffsetArbitrary.java
```diff
@@ -0,0 +1,39 @@
...
+		if (min.getTotalSeconds() > max.getTotalSeconds()) {
+			return atTheEarliest(max).atTheLatest(min);
+		}
+		return atTheEarliest(min).atTheLatest(max);
+	}
+
+	/**
+	 * Set the allowed lower {@code min} (included) bounder of generated zone offset values.
+	 */
+	ZoneOffsetArbitrary atTheEarliest(ZoneOffset min);
```
jlink, line 32: not worthwhile

time/src/main/java/net/jqwik/time/api/arbitraries/ZoneOffsetArbitrary.java
```diff
@@ -0,0 +1,39 @@
...
+
+	/**
+	 * Set the allowed upper {@code max} (included) bounder of generated zone offset values.
+	 */
+	ZoneOffsetArbitrary atTheLatest(ZoneOffset max);
```
jlink, line 37: not worthwhile

time/src/main/java/net/jqwik/time/internal/properties/arbitraries/DefaultOffsetTimeArbitrary.java
```diff
@@ -0,0 +1,80 @@
...
+import net.jqwik.api.arbitraries.*;
+import net.jqwik.time.api.*;
+import net.jqwik.time.api.arbitraries.*;
+
+import static org.apiguardian.api.API.Status.*;
+
+@API(status = INTERNAL)
+public class DefaultOffsetTimeArbitrary extends ArbitraryDecorator<OffsetTime> implements OffsetTimeArbitrary {
+
+	private LocalTimeArbitrary localTimes;
```
jlink, line 18: Yes, that's how imagined embedded arbitraries!

time/src/main/java/net/jqwik/time/internal/properties/arbitraries/DefaultZoneOffsetArbitrary.java
```diff
@@ -0,0 +1,157 @@
...
+	private int calculateMinuteValue(ZoneOffset offset) {
+		return Math.abs((offset.getTotalSeconds() % 3600) / 60);
+	}
+
+	private int calculateSecondValue(ZoneOffset offset) {
+		return Math.abs(offset.getTotalSeconds() % 60);
+	}
+
+	@Override
+	public ZoneOffsetArbitrary atTheEarliest(ZoneOffset min) {
```
jlink, line 128: not worth it

time/src/main/java/net/jqwik/time/internal/properties/arbitraries/DefaultZoneOffsetArbitrary.java
```diff
@@ -0,0 +1,157 @@
...
+			throw new IllegalArgumentException("Minimum offset must not be after maximum offset");
+		}
+
+		DefaultZoneOffsetArbitrary clone = typedClone();
+		clone.offsetMin = min;
+		return clone;
+	}
+
+	@Override
+	public ZoneOffsetArbitrary atTheLatest(ZoneOffset max) {
```
jlink, line 143: not worth it

time/src/main/java/net/jqwik/time/internal/properties/providers/DurationArbitraryProvider.java
```diff
@@ -0,0 +1,24 @@
...
+	@Override
+	public boolean canProvideFor(TypeUsage targetType) {
+		return targetType.isAssignableFrom(Duration.class);
+	}
+
+	@Override
+	public Set<Arbitrary<?>> provideFor(TypeUsage targetType, SubtypeProvider subtypeProvider) {
+		if (targetType.isAssignableFrom(Duration.class)) {
+			return Collections.singleton(Times.durations());
+		} else {
```
jlink, line 20: The if condition should not be necessary bc canProvideFor already prevents it.

time/src/main/java/net/jqwik/time/internal/properties/providers/OffsetTimesArbitraryProvider.java
```diff
@@ -0,0 +1,24 @@
...
+	@Override
+	public boolean canProvideFor(TypeUsage targetType) {
+		return targetType.isAssignableFrom(OffsetTime.class);
+	}
+
+	@Override
+	public Set<Arbitrary<?>> provideFor(TypeUsage targetType, SubtypeProvider subtypeProvider) {
+		if (targetType.isAssignableFrom(OffsetTime.class)) {
+			return Collections.singleton(Times.offsetTimes());
+		} else {
```
jlink, line 20: The if condition should not be necessary bc canProvideFor already prevents it.

time/src/main/java/net/jqwik/time/internal/properties/providers/TimeZoneArbitraryProvider.java
```diff
@@ -0,0 +1,23 @@
...
+
+public class TimeZoneArbitraryProvider implements ArbitraryProvider {
+	@Override
+	public boolean canProvideFor(TypeUsage targetType) {
+		return targetType.isAssignableFrom(TimeZone.class);
+	}
+
+	@Override
+	public Set<Arbitrary<?>> provideFor(TypeUsage targetType, SubtypeProvider subtypeProvider) {
+		if (targetType.isAssignableFrom(TimeZone.class)) {
```
jlink, line 17: The if condition should not be necessary bc canProvideFor already prevents it.

time/src/main/java/net/jqwik/time/internal/properties/providers/TimesArbitraryProvider.java
```diff
@@ -0,0 +1,24 @@
...
+
+public class TimesArbitraryProvider implements ArbitraryProvider {
+	@Override
+	public boolean canProvideFor(TypeUsage targetType) {
+		return targetType.isAssignableFrom(LocalTime.class);
+	}
+
+	@Override
+	public Set<Arbitrary<?>> provideFor(TypeUsage targetType, SubtypeProvider subtypeProvider) {
+		if (targetType.isAssignableFrom(LocalTime.class)) {
```
jlink, line 18: The if condition should not be necessary bc canProvideFor already prevents it.

time/src/main/java/net/jqwik/time/internal/properties/providers/ZoneIdArbitraryProvider.java
```diff
@@ -0,0 +1,24 @@
...
+
+public class ZoneIdArbitraryProvider implements ArbitraryProvider {
+	@Override
+	public boolean canProvideFor(TypeUsage targetType) {
+		return targetType.isAssignableFrom(ZoneId.class);
+	}
+
+	@Override
+	public Set<Arbitrary<?>> provideFor(TypeUsage targetType, SubtypeProvider subtypeProvider) {
+		if (targetType.isAssignableFrom(ZoneId.class)) {
```
jlink, line 18: The if condition should not be necessary bc canProvideFor already prevents it.

time/src/main/java/net/jqwik/time/internal/properties/providers/ZoneOffsetsArbitraryProvider.java
```diff
@@ -0,0 +1,24 @@
...
+
+public class ZoneOffsetsArbitraryProvider implements ArbitraryProvider {
+	@Override
+	public boolean canProvideFor(TypeUsage targetType) {
+		return targetType.isAssignableFrom(ZoneOffset.class);
+	}
+
+	@Override
+	public Set<Arbitrary<?>> provideFor(TypeUsage targetType, SubtypeProvider subtypeProvider) {
+		if (targetType.isAssignableFrom(ZoneOffset.class)) {
```
jlink, line 18: The if condition should not be necessary bc canProvideFor already prevents it.

time/src/test/java/net/jqwik/time/api/DurationTests.java
```diff
@@ -0,0 +1,594 @@
...
+		@Property
+		void validDurationIsGenerated(@ForAll("durations") Duration duration) {
+			assertThat(duration).isNotNull();
+		}
+
+		@Group
+		class WorstCases {
+
+			@Property
+			void worstCaseGeneration(
```
jlink, line 38: Numbered test cases are an anti-pattern. Can the numbers be replaced by a description.

time/src/test/java/net/jqwik/time/api/DurationTests.java
```diff
@@ -0,0 +1,594 @@
...
+			DurationArbitrary durations = Times.durations();
+			TestingFalsifier<Duration> falsifier = duration -> duration.compareTo(Duration.ofSeconds(-999_392_192, 709_938_291)) > 0;
+			Duration value = falsifyThenShrink(durations, random, falsifier);
+			assertThat(value).isEqualTo(Duration.ofSeconds(-999_392_192));
+		}
+
+	}
+
+	@Group
+	class ExhaustiveGeneration {
```
jlink, line 296: Exhaustive generation is especially interesting for reduced precision, e.g. to minutes. Those are IMO the more interesting tests. Also numbered tests again.

time/src/test/java/net/jqwik/time/api/LocalTimeTests.java
```diff
@@ -0,0 +1,1335 @@
...
+
+		@Provide
+		Arbitrary<ZoneId> zoneIds() {
+			return Times.zoneIds();
+		}
+
+	}
+
+	@Group
+	class SimpleAnnotations {
```
jlink, line 108: Why is it called SimpleAnnotations? I don't see any special annotations here.

jlink (commented):
What's missing are the annotations, e.g. TimeRange ZoneRange

jlink:
The review is not complete. The chunk of changes is just too big. We should
strive for smaller chunks in the future. I'll do another review when you've
tackled the open points.

Am Fr., 19. Feb. 2021 um 16:59 Uhr schrieb Johannes Link <
business@johanneslink.net>:

time/src/main/java/net/jqwik/time/api/Times.java
zinki97: I don't think so. 
`zoneIds()` generates 600 values,
`timeZones()` generates 628 values.

time/src/test/java/net/jqwik/time/api/DurationTests.java
zinki97: There isn't the method ofPrecision at the moment. Durations only have two methods: `getSeconds()` and `getNanos()` so I thought it is not necessary. Should I add it?

zinki97: I have the between test cases at ExhaustiveGeneration because I want to know if really all values can be generated.

jlink: Yes, I think DurationArbitrary should get an ofPrecision()

zinki97: Okay I'll work on it. If you'd like, you can proceed with your review while I work on ofPrecision.

time/src/main/java/net/jqwik/time/internal/properties/arbitraries/DefaultZoneOffsetArbitrary.java
```diff
@@ -0,0 +1,147 @@
...
+
+	@Override
+	public ZoneOffsetArbitrary between(ZoneOffset min, ZoneOffset max) {
+		if (min.getTotalSeconds() > max.getTotalSeconds()) {
+			ZoneOffset remember = min;
+			min = max;
+			max = remember;
+		}
+
+		if (min.getTotalSeconds() < DEFAULT_MIN.getTotalSeconds() || min.getTotalSeconds() > DEFAULT_MAX
```
jlink, line 135: Strange formatting. I'd rather have something like:
```java
		if (min.getTotalSeconds() < DEFAULT_MIN.getTotalSeconds() 
					|| min.getTotalSeconds() > DEFAULT_MAX.getTotalSeconds() 
					|| max.getTotalSeconds() > DEFAULT_MAX.getTotalSeconds()) {
```

jlink (commented):
The tests are very thorough!

If you've tackled the outstanding comments, I think it's time for merging the PR.

time/src/main/java/net/jqwik/time/internal/properties/arbitraries/DefaultZoneOffsetArbitrary.java
zinki97: Oh yes, this is the work of "Reformat Code"

zinki97:
Tomorrow I'll start with the annotations.

time/src/main/java/net/jqwik/time/internal/properties/arbitraries/DefaultZoneOffsetArbitrary.java
jlink: Reformat Code is not perfect.

zinki97:
*Request review *

time/src/main/java/net/jqwik/time/internal/properties/configurators/MonthDayRangeConfigurator.java
```diff
@@ -16,12 +16,8 @@ protected boolean acceptTargetType(TypeUsage targetType) {
 	}
 
 	public Arbitrary<?> configure(Arbitrary<?> arbitrary, MonthDayRange range) {
-		if (arbitrary instanceof MonthDayArbitrary) {
-			MonthDayArbitrary monthDayArbitrary = (MonthDayArbitrary) arbitrary;
-			return monthDayArbitrary.between(isoDateToMonthDay(range.min()), isoDateToMonthDay(range.max()));
-		} else {
-			return arbitrary;
-		}
+		MonthDayArbitrary monthDayArbitrary = (MonthDayArbitrary) arbitrary;
```
jlink, line 19: see above

time/src/main/java/net/jqwik/time/internal/properties/configurators/MonthRangeConfigurator.java
```diff
@@ -33,11 +33,9 @@ protected boolean acceptTargetType(TypeUsage targetType) {
 		} else if (arbitrary instanceof YearMonthArbitrary) {
 			YearMonthArbitrary yearMonthArbitrary = (YearMonthArbitrary) arbitrary;
 			return yearMonthArbitrary.monthBetween(range.min(), range.max());
-		} else if (arbitrary instanceof MonthDayArbitrary) {
+		} else {
```
jlink, line 36: see above

time/src/main/java/net/jqwik/time/internal/properties/configurators/OffsetRangeConfigurator.java
```diff
@@ -0,0 +1,28 @@
...
+	@Override
+	protected boolean acceptTargetType(TypeUsage targetType) {
+		return targetType.isAssignableFrom(ZoneOffset.class) || targetType.isAssignableFrom(OffsetTime.class);
+	}
+
+	public Arbitrary<?> configure(Arbitrary<?> arbitrary, OffsetRange range) {
+		if (arbitrary instanceof ZoneOffsetArbitrary) {
+			ZoneOffsetArbitrary zoneOffsetArbitrary = (ZoneOffsetArbitrary) arbitrary;
+			return zoneOffsetArbitrary.between(ZoneOffset.of(range.min()), ZoneOffset.of(range.max()));
+		} else {
```
jlink, line 22: see above.

time/src/main/java/net/jqwik/time/internal/properties/configurators/PrecisionConfigurator.java
```diff
@@ -0,0 +1,32 @@
...
+	}
+
+	public Arbitrary<?> configure(Arbitrary<?> arbitrary, Precision range) {
+		if (arbitrary instanceof LocalTimeArbitrary) {
+			LocalTimeArbitrary localTimeArbitrary = (LocalTimeArbitrary) arbitrary;
+			return localTimeArbitrary.ofPrecision(range.ofPrecision());
+		} else if (arbitrary instanceof OffsetTimeArbitrary) {
+			OffsetTimeArbitrary offsetTimeArbitrary = (OffsetTimeArbitrary) arbitrary;
+			return offsetTimeArbitrary.ofPrecision(range.ofPrecision());
+		} else {
```
jlink, line 27: see above

time/src/main/java/net/jqwik/time/internal/properties/configurators/SecondRangeConfigurator.java
```diff
@@ -0,0 +1,27 @@
...
+	@Override
+	protected boolean acceptTargetType(TypeUsage targetType) {
+		return targetType.isAssignableFrom(LocalTime.class) || targetType.isAssignableFrom(OffsetTime.class);
+	}
+
+	public Arbitrary<?> configure(Arbitrary<?> arbitrary, SecondRange range) {
+		if (arbitrary instanceof LocalTimeArbitrary) {
+			LocalTimeArbitrary localTimeArbitrary = (LocalTimeArbitrary) arbitrary;
+			return localTimeArbitrary.secondBetween(range.min(), range.max());
+		} else {
```
jlink, line 22: see above

time/src/main/java/net/jqwik/time/internal/properties/configurators/TimeRangeConfigurator.java
```diff
@@ -0,0 +1,32 @@
...
+	@Override
+	protected boolean acceptTargetType(TypeUsage targetType) {
+		return targetType.isAssignableFrom(LocalTime.class) || targetType.isAssignableFrom(OffsetTime.class);
+	}
+
+	public Arbitrary<?> configure(Arbitrary<?> arbitrary, TimeRange range) {
+		if (arbitrary instanceof LocalTimeArbitrary) {
+			LocalTimeArbitrary localTimeArbitrary = (LocalTimeArbitrary) arbitrary;
+			return localTimeArbitrary.between(stringToLocalTime(range.min()), stringToLocalTime(range.max()));
+		} else {
```
jlink, line 22: see above

time/src/main/java/net/jqwik/time/internal/properties/configurators/YearMonthRangeConfigurator.java
```diff
@@ -16,12 +16,8 @@ protected boolean acceptTargetType(TypeUsage targetType) {
 	}
 
 	public Arbitrary<?> configure(Arbitrary<?> arbitrary, YearMonthRange range) {
-		if (arbitrary instanceof YearMonthArbitrary) {
-			YearMonthArbitrary yearMonthArbitrary = (YearMonthArbitrary) arbitrary;
-			return yearMonthArbitrary.between(isoDateToYearMonth(range.min()), isoDateToYearMonth(range.max()));
-		} else {
-			return arbitrary;
-		}
+		YearMonthArbitrary yearMonthArbitrary = (YearMonthArbitrary) arbitrary;
```
jlink, line 19: if you really want to restrict it to YearMonthArbitrary instances the `arbitrary` parameter can be typed accordingly. This will prevent casting error on wrong usage.

time/src/main/java/net/jqwik/time/internal/properties/configurators/YearRangeConfigurator.java
```diff
@@ -33,11 +33,9 @@ protected boolean acceptTargetType(TypeUsage targetType) {
 		} else if (arbitrary instanceof YearMonthArbitrary) {
 			YearMonthArbitrary yearMonthArbitrary = (YearMonthArbitrary) arbitrary;
 			return yearMonthArbitrary.yearBetween(range.min(), range.max());
-		} else if (arbitrary instanceof YearArbitrary) {
+		} else {
```
jlink, line 36: see above

jlink (changes requested):
Unless I've overlooked something, removing handling of unknown arbitrary types in configurators was wrong. Maybe coverage analysis lead you astray. There were probably tests missing to check for decent behaviour when time-related annotations are used on non-time related parameters.

time/src/main/java/net/jqwik/time/internal/properties/configurators/DateRangeConfigurator.java
```diff
@@ -26,11 +26,9 @@ protected boolean acceptTargetType(TypeUsage targetType) {
 		} else if (arbitrary instanceof CalendarArbitrary) {
 			CalendarArbitrary calendarArbitrary = (CalendarArbitrary) arbitrary;
 			return calendarArbitrary.between(isoDateToCalendar(range.min()), isoDateToCalendar(range.max()));
-		} else if (arbitrary instanceof DateArbitrary) {
+		} else {
```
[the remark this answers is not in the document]
zinki97: Isn't it protected by `acceptTargetType`?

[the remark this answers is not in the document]
jlink: You can create any target type in diverse manners, e.g, Arbitraries.just(LocalTime.now())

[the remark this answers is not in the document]
zinki97: Ah okay. I thought it was unnecessary code.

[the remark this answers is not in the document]
zinki97: I have new testcases and it fails as you expect.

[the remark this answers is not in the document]
jlink: I’d say the tests are missing. In the case where none of the specialised Arbitraries are present a generic filter could be used.

time/src/main/java/net/jqwik/time/internal/properties/arbitraries/DefaultDurationArbitrary.java
```diff
@@ -0,0 +1,237 @@
...
+		Arbitrary<BigInteger> bigIntegers = Arbitraries.bigIntegers()
+													   .between(min, max)
+													   .withDistribution(RandomDistribution.uniform())
+													   .edgeCases(edgeCases -> edgeCases.includeOnly(min, BigInteger.ZERO, max));
+
+		return bigIntegers.map(this::calculateDuration);
+
+	}
+
+	private Duration calculateEffectiveMin() {
```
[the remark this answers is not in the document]
zinki97: okay

time/src/test/java/net/jqwik/time/api/TimesConstraintTests.java
```diff
@@ -0,0 +1,1104 @@
+package net.jqwik.time.api;
```
jlink, line 1: I noticed some System.out.println in the tests. They pollute the output.

time/src/test/java/net/jqwik/time/api/TimesConstraintTests.java
```diff
@@ -0,0 +1,1104 @@
...
+@Group
+public class TimesConstraintTests {
+
+	@Group
+	class LocalTimeConstraints {
+
+		@Group
+		class Constraints {
+
+			@Property
```
jlink, line 25: There's a test missing (in other places as well) along the following lines:
```java
@Property
void timeRangeForProvidedTime(@ForAll("myTimes") @TimeRange(min = "00:01:00", max = "00:02:00") LocalTime time) {
	assertThat(time).isBetween(
			LocalTime.of(0, 1, 0),
			LocalTime.of(0, 2, 0)
	);
}

@Provide
Arbitrary<LocalTime> myTimes() {
	return Arbitraries.of(
			LocalTime.of(0, 0, 1),
			LocalTime.of(0, 1, 1),
			LocalTime.of(0, 2, 1)
	);
}
```

The point is to allow the constraints to work even when a self-made arbitrary generates the time (or whatever). 

Implementationwise this can be done by a simple filter in those cases. And then you probably want to split `TimeRangeConfigurator` in two classes: `TimeRangeForLocalTimeConfigurator` and `TimeRangeForOffsetTimeConfigurator` in order to minimize the number of `instanceof` usages.

And this holds for most other time-related constraint annotations as well I'm afraid.

jlink (commented):
After tackling the two points mentioned I think the PR is ready for merging.

time/src/test/java/net/jqwik/time/api/TimesConstraintTests.java
zinki97: I forgot to remove it

time/src/test/java/net/jqwik/time/api/TimesConstraintTests.java
zinki97: How should I handle Periods? 

The problem with periods: For example a Period can have 3 Years, 293 Months and 211 days. How can I compare it with a period with for example 4 Years, 2 Month and 11 days?

jlink: I don't see the relation to my comment above, but anyway: you could compare by effective length. Or maybe a period range does not make sense.

zinki97: My comment is here because I'll change all configurators so that they work for user arbitraries.

zinki97: Done.

zinki97:
Why the tests fail?

jlink:
> Why the tests fail?

I guess the arbitrary parameter should be of type `Arbitrary<Integer>`

jlink (commented):
I think all configurators for the same annotation could reside as inner static classes of one collecting class, e.g.

```
class DayOfMonthRangeConfigurator
    static class ForDate
    static class ForLocalDate

```

There might even be a chance of code sharing among them.

zinki97:
> > Why the tests fail?
> 
> I guess the arbitrary parameter should be of type `Arbitrary<Integer>`

Sorry I mean the build not the tests

jlink:
> > > Why the tests fail?
> > 
> > 
> > I guess the arbitrary parameter should be of type `Arbitrary<Integer>`
> 
> Sorry I mean the build not the tests

My suggestion stands, but I haven't tried it

zinki97:
> > > > Why the tests fail?
> > > 
> > > 
> > > I guess the arbitrary parameter should be of type `Arbitrary<Integer>`
> > 
> > 
> > Sorry I mean the build not the tests
> 
> My suggestion stands, but I haven't tried it

On my machine it works. Your last push failed, too.

jlink:
> > > > > Why the tests fail?
> > > > 
> > > > 
> > > > I guess the arbitrary parameter should be of type `Arbitrary<Integer>`
> > > 
> > > 
> > > Sorry I mean the build not the tests
> > 
> > 
> > My suggestion stands, but I haven't tried it
> 
> On my machine it works. Your last push failed, too.

For different reasons though.

zinki97:
> > > > > > Why the tests fail?
> > > > > 
> > > > > 
> > > > > I guess the arbitrary parameter should be of type `Arbitrary<Integer>`
> > > > 
> > > > 
> > > > Sorry I mean the build not the tests
> > > 
> > > 
> > > My suggestion stands, but I haven't tried it
> > 
> > 
> > On my machine it works. Your last push failed, too.
> 
> For different reasons though.

I don't know I saw it only on my mobile phone. I'll try your ide tomorrow.

jlink:
> > > > > > > Why the tests fail?
> > > > > > 
> > > > > > 
> > > > > > I guess the arbitrary parameter should be of type `Arbitrary<Integer>`
> > > > > 
> > > > > 
> > > > > Sorry I mean the build not the tests
> > > > 
> > > > 
> > > > My suggestion stands, but I haven't tried it
> > > 
> > > 
> > > On my machine it works. Your last push failed, too.
> > 
> > 
> > For different reasons though.
> 
> I don't know I saw it only on my mobile phone. I'll try your ide tomorrow.

A clean build with the exact same JDK as github actions is using (8.0.282-zulu) produced the same error locally.

zinki97:
> I think all configurators for the same annotation could reside as inner static classes of one collecting class, e.g.
> 
> ```
> class DayOfMonthRangeConfigurator
>     static class ForDate
>     static class ForLocalDate
> ```
> 
> There might even be a chance of code sharing among them.

Done.

Ready for review

jlink:
@zinki97 Good work. Coverage is even better than in jqwik core!

I'll merge it soon.

jlink (approved):

Issue #175, opened 2021-03-09
Add Generation of DateTimes


jlink:
## Testing Problem

Now that generation of dates (https://github.com/jlink/jqwik/issues/140) and generation of times (https://github.com/jlink/jqwik/issues/154) is done, the last step is to add generation of DateTimes.

## Suggested Solution

Add the capability to the "jqwik-time" module to generate date-time values.

Foundation for generation are the types in java.time. Mappers to "old" JDK classes like Date, Calendar etc. should be provided for convenience.

Add a new class to the base package `net.jqwik.api.time`: `net.jqwik.time.api.DateTimes`

jlink:
@zinki97 You might want to start with an API suggestion

zinki97:
Design idea: **WORK IN PROGRESS**

zinki97:
@jlink What do you think about the design idea?

zinki97:
> To be consisten with class `Dates` I suggest the following changes:
> 
> * `localDateTimes()` -> `dateTimes()`
> * `dates()` -> `dateTimesAsDate()`
> * calendars() -> `dateTimesAsCalendar()`
> 
> I hopen that `datesAsDate()` and `datesAsCalendar()` could "just" use `dateTimesAsDates()` and `dateTimesAsCalendar()` with precision set to `DAYS`.

You're right, it is better. Then let's start with the implementation.

zinki97:
Should LocalDateTime include `between(LocalDateTime, LocalDateTime)` or only `betweenDate(LocalDate, LocalDate)` and `betweenTime(LocalTime, LocalTime)`?

The second solution is better for delegation, but the first solution is possible, too.

zinki97:
Which idea is the best? @jlink 

idea 1) is possible but bad for delegation, 

I vote for idea 3).

Edit: 
idea 1)
`between(LocalDateTime, LocalDateTime)`
`atTheEarliest(LocalDateTime)`
`atTheLatest(LocalDateTime)`
`dateBetween(LocalDate, LocalDate)`
`timeBetween(LocalTime, LocalTime)`

idea 2)
`dateBetween(LocalDate, LocalDate)`
`timeBetween(LocalTime, LocalTime)`

idea 3):
`between(LocalDate, LocalDate)`
`atTheEarliest(LocalDate)`
`atTheLatest(LocalDate)`
`timeBetween(LocalTime, LocalTime)`

jlink:
[...]

PR #178, opened 2021-03-29
Generation of DateTimes 2) Date based methods

zinki97:

In this step the first date-based methods are implemented


time/src/main/java/net/jqwik/time/internal/properties/arbitraries/DefaultLocalDateTimeArbitrary.java
```diff
@@ -21,14 +21,21 @@
 	private LocalDateTime min = null;
 	private LocalDateTime max = null;
 
+	private LocalDate minDate = null;
```
jlink, line 24: There's so much duplication with LocalDate and LocalTime arbitrary implementations. I wonder if there's a way to get rid of that.

zinki97: I don't think so because you need it for the calculation of effectiveMin/effectiveMax which is different from the calculation in the time/date class.

Some things (i.e. monthMin/Max) you can exclude because you don't need them for the calculation but then it ist not consistent. Should I exclude these things?

jlink: I’m not saying simple delegation is possible but there should be a way to put all the duplicated logic of checking validity and creating error messages/exceptions in a common (new) abstraction.

zinki97: To make sure that I understand you right:
Do you mean a class like the new class LocalDateTimeSupport?

Do you have a better name for it?
In which package is the class correct?

jlink: I was thinking of methods like `yearBetween()` and `monthBetween()` etc. They should be the same (structurally and as of error/special case logic). But if you compare `DefaultLocalDateArbitrary.yearBetween()` with `DefaultLocalDateTimeArbitrary.yearBetween()` you will see that they behave differently. Unifying both in s.t. like a `YearBetween` class could do two things: 1) Get rid of duplicated code 2) Show where the logic is different although it should be the same.

jlink: > Unifying both in s.t. like a `YearBetween` class could do two things: 1) Get rid of duplicated code 2) Show where the logic is different although it should be the same.

I'm not sure this is the best design approach but in my head there's a small hierarchy looming

```
ValueRange<T, U>
    YearBetween<Integer, LocalDate>
    MonthBetween<Integer, LocalDate>
    DayOfMonthBetween<Integer, Integer>
...
```

That's just a sketch, of course.

zinki97: I think it is not completly like you expected but it works.

zinki97:
Has the failing test something to do with my code?

jlink:
> Has the failing test something to do with my code?

For you to find out ;-)

time/src/main/java/net/jqwik/time/internal/properties/arbitraries/valueRanges/ValueRange.java
```diff
@@ -0,0 +1,61 @@
+package net.jqwik.time.internal.properties.arbitraries.valueRanges;
+
+public abstract class ValueRange<T> {
```
jlink, line 3: For me the name is too generic for its current usage. It's only used for "Betweens" so I'd call it like that.

time/src/main/java/net/jqwik/time/internal/properties/arbitraries/valueRanges/ValueRange.java
```diff
@@ -0,0 +1,61 @@
...
+
+	public T getMin() {
+		return min;
+	}
+
+	public T getMax() {
+		return max;
+	}
+
+	protected class Parameter {
```
jlink, line 37: As far as I can see, instances of Parameter are only used in methods of subclasses. The class doesn't pull its weight IMO. Indeed it makes the interface of minMaxChanger() and exceptionCheck() less obvious. I'd get rid of it and just use the two parameters min and max.

time/src/main/java/net/jqwik/time/internal/properties/arbitraries/valueRanges/ValueRange.java
```diff
@@ -0,0 +1,61 @@
...
+
+	private T min;
+	private T max;
+
+	protected void exceptionCheck(Parameter parameter) {
+		//Override if needed
+		//do nothing in default case
+	}
+
+	protected void minMaxChanger(Parameter parameter) {
```
jlink, line 13: There's only one subclass that does not swap min and max. So swapping could be the default implementation. Moreover, the name could be more telling, e.g. "ensureMinIsBeforeMax" or something similar.

time/src/main/java/net/jqwik/time/internal/properties/arbitraries/valueRanges/ValueRange.java
```diff
@@ -0,0 +1,61 @@
+package net.jqwik.time.internal.properties.arbitraries.valueRanges;
+
+public abstract class ValueRange<T> {
+
+	private T min;
+	private T max;
+
+	protected void exceptionCheck(Parameter parameter) {
```
jlink, line 8: "exceptionCheck" is a name that suggests that an exception is being checked. What you really do is checking the range's validity. Suggested name: checkRangeIsValid or checkValidity or st like that.

time/src/main/java/net/jqwik/time/internal/properties/arbitraries/valueRanges/LocalDateTimeBetween.java
```diff
@@ -0,0 +1,18 @@
+package net.jqwik.time.internal.properties.arbitraries.valueRanges;
+
+import java.time.*;
+
+public class LocalDateTimeBetween extends ValueRange<LocalDateTime> {
```
jlink, line 5: Why does it NOT swap min and max automatically if min > max?

time/src/main/java/net/jqwik/time/internal/properties/arbitraries/valueRanges/LocalDateBetween.java
```diff
@@ -0,0 +1,26 @@
+package net.jqwik.time.internal.properties.arbitraries.valueRanges;
+
+import java.time.*;
+
+public class LocalDateBetween extends ValueRange<LocalDate> {
```
jlink, line 5: Why does it NOT swap min and max automatically if min > max?

time/src/main/java/net/jqwik/time/internal/properties/arbitraries/valueRanges/LocalTimeBetween.java
```diff
@@ -0,0 +1,14 @@
+package net.jqwik.time.internal.properties.arbitraries.valueRanges;
+
+import java.time.*;
+
+public class LocalTimeBetween extends ValueRange<LocalTime> {
```
jlink, line 5: Why does it NOT swap min and max automatically if min > max?

time/src/main/java/net/jqwik/time/internal/properties/arbitraries/valueRanges/YearMonthBetween.java
```diff
@@ -0,0 +1,26 @@
+package net.jqwik.time.internal.properties.arbitraries.valueRanges;
+
+import java.time.*;
+
+public class YearMonthBetween extends ValueRange<YearMonth> {
```
jlink, line 5: Why does it NOT swap min and max automatically if min > max?

time/src/main/java/net/jqwik/time/internal/properties/arbitraries/valueRanges/ZoneOffsetBetween.java
```diff
@@ -0,0 +1,12 @@
+package net.jqwik.time.internal.properties.arbitraries.valueRanges;
+
+import java.time.*;
+
+public class ZoneOffsetBetween extends ValueRange<ZoneOffset> {
+	@Override
+	protected void minMaxChanger(Parameter parameter) {
+		if (parameter.getMin().getTotalSeconds() > parameter.getMax().getTotalSeconds()) {
```
jlink, line 8: Does comparing ZoneOffset instances not work here?

zinki97:
> > Has the failing test something to do with my code?
> 
> For you to find out ;-)

I've run the test over 2,000 times and it hasn't failed again.

jlink (commented):
Looks good. Is it ready for merging?

IMO all that's missing is default generation for Calendar and Date. Will that go in another PR?

zinki97:
> Looks good. Is it ready for merging?
> 
> IMO all that's missing is default generation for Calendar and Date. Will that go in another PR?

In order to have smaller pull requests this PR is ready for merging.

What is currently missing:
- some date based methods (leapYears, ...)
- time based methods
- Calendar
- Date

But I want to put these things in four more pull requests (each point a PR)

PR #184, opened 2021-04-14
Generation of DateTimes 4) Time based methods

zinki97:
See #175

In this step all time-based methods are implemented


zinki97:
Have you changed something on the code style? @jlink

jlink:
Maybe two weeks ago I aligned .editorconfig with the IntelliJ specific xml configuration. Any special thing you noticed?

zinki97:
> Maybe two weeks ago I aligned .editorconfig with the IntelliJ specific xml configuration. Any special thing you noticed?

Only that about 400 lines were changed in `LocalDateTimeTests`.

jlink:
> Only that about 400 lines were changed in `LocalDateTimeTests`.

which cleanly shows that the class is way too long.

zinki97:
> > Only that about 400 lines were changed in `LocalDateTimeTests`.
> 
> which cleanly shows that the class is way too long.

What do you suggest? Packages with smaller classes?

zinki97:
> > > Only that about 400 lines were changed in `LocalDateTimeTests`.
> > 
> > 
> > which cleanly shows that the class is way too long.
> 
> What do you suggest? Packages with smaller classes?

My suggest is to make smaller classes in their own packages. (At the end oft this step) For example package `LocalDateTimeTests` with class `ExhaustiveGeneration`, ...

What do you think @jlink ?

jlink:
> My suggest is to make smaller classes in their own packages. (At the end oft this step) For example package `LocalDateTimeTests` with class `ExhaustiveGeneration`, ...
> 
> What do you think @jlink ?

Sounds reasonable. Pay attention to naming, not all names are picked up by JUnit by default. Stick with Java conventions: package names start lower case. 
What about „net.jqwik.time.localDateTime.ExhaustiveGenerationTests“

zinki97:
> > My suggest is to make smaller classes in their own packages. (At the end oft this step) For example package `LocalDateTimeTests` with class `ExhaustiveGeneration`, ...
> > What do you think @jlink ?
> 
> Sounds reasonable. Pay attention to naming, not all names are picked up by JUnit by default. Stick with Java conventions: package names start lower case.
> What about „net.jqwik.time.localDateTime.ExhaustiveGenerationTests“

Thanks for the hint. Sounds good!

zinki97:
`effectiveDate = date.plusDays(1);`

I think this line will cause an error if year/month/dayOfMonthBetween/etc. is set ://

zinki97:
I also found a bug in `DefaultLocalTimeArbitrary`

jlink:
> I also found a bug in `DefaultLocalTimeArbitrary`

More info? Failing test?

zinki97:
> > I also found a bug in `DefaultLocalTimeArbitrary`
> 
> More info? Failing test?

The failing test was a new test in `LocalDateTimeTests`. After wondering why it wasn't working I found the problem in `DefaultLocalTimeArbitrary` after experimenting a bit. This shows me that there a soreme test cases missing.

Correct calculation of effective maximum time sometimes does not work if you set hours, minutes, seconds and a max time.

Example:
```
max-time: 13:25:31
hourBetween: 11, 11
minuteBetween: 22, 22
secondBetween: 31, 34
```

It calculates the max effective time `11:22:31`

jlink:
> max-time: 13:25:31
> hourBetween: 11, 11
> minuteBetween: 22, 22
> secondBetween: 31, 34
> It calculates the max effective time `11:22:31` instead of `11:22:34`

That’s subtle.

time/src/main/java/net/jqwik/time/internal/properties/arbitraries/DefaultLocalTimeArbitrary.java
```diff
@@ -139,15 +141,24 @@ public static LocalTime calculateEffectiveMinWithPrecision(LocalTime effective,
...
+			effective = effective.withNano(999_999_999);
 			if (minuteBetween.getMax() < effective.getMinute()) {
 				effective = effective.withMinute(minuteBetween.getMax());
+				effective = effective.withSecond(59);
 				if (secondBetween.getMax() < effective.getSecond()) {
 					effective = effective.withSecond(secondBetween.getMax());
-					effective = effective.withNano(999_999_999);
 				}
 			}
 		}
```
zinki97, line 164: I think I fixed the problem, but I will write more test cases.

zinki97: The problem is also present in calculating min values....

zinki97: It should work now

zinki97:
In the class `DefaultLocalDateArbitrary` filter prevents the bug. But in some cases edge cases are not generated.

time/src/main/java/net/jqwik/time/internal/properties/arbitraries/DefaultLocalTimeArbitrary.java
zinki97: There was still a bug, it has now been fixed

zinki97:
> `effectiveDate = date.plusDays(1);`
> 
> I think this line will cause an error if year/month/dayOfMonthBetween/etc. is set ://

I thought about it: It calculates a wrong effective min date when month or dayOfMonth is restricted. But it doesn't matter. It don't affect exhaustive generation or edge cases. I'll add some tests to verify my thoughts.

I think a real min date is not necessary in this case. The LocalDateArbitrary class calculates the real min date in a later step.

time/src/main/java/net/jqwik/time/internal/properties/arbitraries/DefaultLocalDateArbitrary.java
```diff
@@ -89,11 +112,19 @@ private int latestAllowedMonth() {
...
+		}
 		if (earliestMonth > effective.getMonth().getValue()) {
-			return effective.withMonth(earliestMonth);
+			effective = effective.withDayOfMonth(1).withMonth(earliestMonth);
 		}
 		if (dayOfMonthBetween.getMin() != null && dayOfMonthBetween.getMin() > effective.getDayOfMonth()) {
-			return effective.withDayOfMonth(dayOfMonthBetween.getMin());
+			try {
+				effective = effective.withDayOfMonth(dayOfMonthBetween.getMin());
+			} catch (DateTimeException e) {
```
jlink, line 125: Try-catch as case differentiation is a smell. Can you instead test for the condition that will lead to the exception?
If that is too much work, at least document why the exception can occur here.

zinki97: Done

time/src/main/java/net/jqwik/time/internal/properties/arbitraries/DefaultLocalDateArbitrary.java
```diff
@@ -70,21 +70,74 @@
...
+						&& allowedMonths.get().contains(date.getMonth())
+				) {
+					return;
+				}
+			}
 		}
 		throw new IllegalArgumentException("These min/max configurations cannot be used together: No values are possible.");
 	}
 
+	private boolean valuesArePossible() {
```
jlink, line 92: This looks terribly complicated. What happens if you just don't check? Won't the generator just fail at first attempt and tell that no value can be generated?

time/src/main/java/net/jqwik/time/internal/properties/arbitraries/DefaultLocalDateArbitrary.java
```diff
@@ -120,15 +121,28 @@ private LocalDate effectiveMinDate() {
 			effective = effective.withDayOfMonth(1).withMonth(earliestMonth);
 		}
 		if (dayOfMonthBetween.getMin() != null && dayOfMonthBetween.getMin() > effective.getDayOfMonth()) {
-			try {
+			if (isValidDate(effective.getYear(), effective.getMonth(), dayOfMonthBetween.getMin())) {
```
jlink, line 180: I know I suggested to get rid of try-catch where appropriate. Here however you try (maybe successfully maybe not) to mimick what effective.withDayOfMonth() does in the library. I don't think that's well invested code.

My suggestion was to get rid of it OR DOCUMENT why an exception can occur here.

time/src/main/java/net/jqwik/time/internal/properties/arbitraries/DefaultLocalDateTimeArbitrary.java
```diff
@@ -101,15 +102,14 @@ private LocalDateTime calculateEffectiveMinDate(LocalDateTime effective) {
...
-			try {
+			if (date.getYear() < 999_999_999 || date.getMonth() != DECEMBER || date.getDayOfMonth() < 31) {
 				effectiveDate = date.plusDays(1);
-			} catch (DateTimeException dateTimeException) {
-				throw e;
+			} else {
+				throw new IllegalArgumentException("Cannot use this min value with precision " + ofPrecision.get());
 			}
 			date = effectiveDate;
 		}
```
jlink, line 115: Let's revisit this code together. IMO it does not communicate well what's happening here.

jlink:
Can you roll back to https://github.com/jlink/jqwik/pull/184/commits/2c70dd4e23c2c9ff46235671f52e0ec07748a232? 
(And keep the two additional commits for a later PR).

Too much has happened. I'd like to get that into main first before proceeding with the new API methods.

zinki97:
After our meeting it is clear what should happen with min value `11:52:29` and precision `MINUTES`. (exception)

But what about the max value `11:52:29` and precision `MINUTES`? Also exception or cut it to 11:52:00?

jlink:
Yes. I think so. 

Von meinem iPhone gesendet

zinki97:
> Yes. I think so.

What do you think? Exception? Or Cut it? :D

jlink:
Exception. Gives us the option to later add an optional rounding parameter. 

Von meinem iPad gesendet

jlink:
> Exception. Gives us the option to later add an optional rounding parameter. Von meinem iPad gesendet

Here's how it's being handled in big decimal generation:

```java
	@Property
	void test2(@ForAll("bigs") BigDecimal bigInteger) {

	}

	@Provide
	Arbitrary<BigDecimal> bigs() {
		return Arbitraries.bigDecimals()
						  .between(new BigDecimal("1.111"), new BigDecimal("2.222"))
						  .ofScale(2);
	}
```

Will throw the following exception on run:

```
net.jqwik.api.JqwikException:
    Decimal value 1.111 cannot be represented with scale 2.
    You may want to use a higher scale
```

I suggest a similiar revealing error message. I'm not quite sure if IllegalArgumentException would be preferrable.

zinki97:
> I suggest a similiar revealing error message. I'm not quite sure if IllegalArgumentException would be preferrable.

The current exception is:

```
java.lang.IllegalArgumentException:
    Can't use 13:59:59.999999999 as maximum time with precision Seconds.
```

Should I change it?

jlink:
> Should I change it?

You could add „You may want to round the time to Seconds or change precision to Nanos“

zinki97:
> You may want to round the time to Seconds or change precision to Nanos

I've changed it to `You may want to round the time to Seconds or change the precision.`

jlink:
> I've changed it to `You may want to round the time to Seconds or change the precision.`

Did you change it or did you add the sentence?

zinki97:
> > I've changed it to `You may want to round the time to Seconds or change the precision.`
> 
> Did you change it or did you add the sentence?

Add the sentence
(next PR)

jlink:
Build warning:
```
/Users/link/Documents/git-projects/jqwik/time/src/main/java/net/jqwik/time/api/constraints/DateRange.java:17: warning - Tag @see: reference not found: LeapYears
/Users/link/Documents/git-projects/jqwik/time/src/main/java/net/jqwik/time/api/constraints/DayOfMonthRange.java:17: warning - Tag @see: reference not found: LeapYears
/Users/link/Documents/git-projects/jqwik/time/src/main/java/net/jqwik/time/api/constraints/DayOfWeekRange.java:18: warning - Tag @see: reference not found: LeapYears
/Users/link/Documents/git-projects/jqwik/time/src/main/java/net/jqwik/time/api/constraints/MonthDayRange.java:18: warning - Tag @see: reference not found: LeapYears
/Users/link/Documents/git-projects/jqwik/time/src/main/java/net/jqwik/time/api/constraints/MonthRange.java:19: warning - Tag @see: reference not found: LeapYears
/Users/link/Documents/git-projects/jqwik/time/src/main/java/net/jqwik/time/api/constraints/YearMonthRange.java:18: warning - Tag @see: reference not found: LeapYears
/Users/link/Documents/git-projects/jqwik/time/src/main/java/net/jqwik/time/api/constraints/YearRange.java:18: warning - Tag @see: reference not found: LeapYears
```

zinki97:
Could you please review this PR?

time/src/main/java/net/jqwik/time/internal/properties/arbitraries/DefaultDurationArbitrary.java
```diff
@@ -51,91 +44,69 @@
...
+				throwException = (effective.getNano() % 1_000) != 0;
+		}
+		if (throwException) {
+			throwDurationAndPrecisionException(effective.toString(), minimum, ofPrecision.get());
+		}
+	}
+
+	private static void throwDurationAndPrecisionException(String val, boolean minimum, ChronoUnit precision) {
+		String minMax = minimum ? "minimum" : "maximum";
+		throw new IllegalArgumentException("Can't use " + val + " as " + minMax + " duration with precision " + precision + ".\nYou may want to round the duration to " + precision + " or change the precision.");
```
jlink, line 87: Use String.format() instead of String concatenation.

time/src/main/java/net/jqwik/time/internal/properties/arbitraries/DefaultDurationArbitrary.java
```diff
@@ -51,91 +44,69 @@
...
+	private Duration calculateEffectiveMax(DurationBetween durationBetween, OfPrecision ofPrecision) {
+		Duration effective = durationBetween.getMax() != null ? durationBetween.getMax() : calculateMaxPossibleValue(ofPrecision);
+		checkValueAndPrecision(effective, ofPrecision, false);
+		return effective;
+	}
+
+	private Duration calculateMaxPossibleValue(OfPrecision ofPrecision) {
+		switch (ofPrecision.get()) {
+			case HOURS:
+				return Duration.ofSeconds((Long.MAX_VALUE / (60 * 60)) * (60 * 60), 0);
```
jlink, line 99: All those calculations could be on OfPrecision objects

time/src/main/java/net/jqwik/time/internal/properties/arbitraries/DefaultDurationArbitrary.java
```diff
@@ -51,91 +44,69 @@
...
-				}
-			}
-			return effective;
-		} catch (ArithmeticException e) {
-			throw new IllegalArgumentException("Max value must be decreased because of precision " + precision + " but results in a " + e.getMessage());
+	private void checkValueAndPrecision(Duration effective, OfPrecision ofPrecision, boolean minimum) {
+		boolean throwException = false;
+		switch (ofPrecision.get()) {
+			case HOURS:
+				throwException = effective.getSeconds() % 3600 != 0;
```
jlink, line 68: All those calculations could be in OfPrecision

jlink (commented):
Some recommendations for class OfPrecision:
- make it a real immutable value object, get rid of those clone().set/setProgrammatically()
- it can do many calculations itself e.g. toSeconds(), compare(ChronoUnit) etc.
   Just look for how you use instances of it. It will help you get rid of the many switches in the using code.

zinki97:
> * make it a real immutable value object, get rid of those clone().set/setProgrammatically()
> * it can do many calculations itself e.g. toSeconds(), compare(ChronoUnit) etc.
>   Just look for how you use instances of it. It will help you get rid of the many switches in the using code.

I don't really understand what you mean, sorry. Could you give me an example or explain it, please?

jlink:
> I don't really understand what you mean, sorry. Could you give me an example or explain it, please?

Look at LocalDateTime for example. You can create an instance of it, but you cannot change its internal state. Comparing two instances will compare on their value. And LocalDateTime come with many methods for calculation, comparing, transformation etc.  Basically all classes in java.time follow that pattern. If OfPrecision would too, my feeling is that quite some code can become more elegant.

time/src/main/java/net/jqwik/time/internal/properties/arbitraries/DefaultDurationArbitrary.java
```diff
@@ -200,7 +200,7 @@ public DurationArbitrary between(Duration min, Duration max) {
 	@Override
 	public DurationArbitrary ofPrecision(ChronoUnit ofPrecision) {
 		DefaultDurationArbitrary clone = typedClone();
-		clone.ofPrecision.set(ofPrecision);
+		clone.ofPrecision = this.ofPrecision.set(ofPrecision);
 		return clone;
 	}
```
zinki97, line 206: Is this what you mean?

time/src/main/java/net/jqwik/time/internal/properties/arbitraries/DefaultDurationArbitrary.java
```diff
@@ -180,7 +180,7 @@ private void setOfPrecisionImplicitly(DefaultDurationArbitrary clone, Duration d
 		}
 		ChronoUnit ofPrecision = DefaultLocalTimeArbitrary.calculateOfPrecisionFromNanos(duration.getNano());
 		if (clone.ofPrecision.get().compareTo(ofPrecision) > 0) {
-			clone.ofPrecision.setProgrammatically(ofPrecision);
+			clone.ofPrecision = this.ofPrecision.setProgrammatically(ofPrecision);
 		}
 	}
```
zinki97, line 185: Is this what you mean?

time/src/main/java/net/jqwik/time/internal/properties/arbitraries/valueRanges/OfPrecision.java
```diff
@@ -11,19 +11,34 @@
...
+	private OfPrecision set(ChronoUnit precision, boolean precisionSet) {
 		if (!ALLOWED_PRECISIONS.contains(precision)) {
 			throw new IllegalArgumentException("Precision value must be one of these ChronoUnit values: HOURS, MINUTES, SECONDS, MILLIS, MICROS, NANOS");
 		}
-		this.precision = precision;
+		if (this.precision.equals(precision) && this.precisionSet == precisionSet) {
+			return this;
+		}
+		return new OfPrecision(precision, precisionSet);
 	}
```
zinki97, line 42: Is this what you mean?

zinki97 (commented):
Is this what you mean? (some examples)

time/src/main/java/net/jqwik/time/internal/properties/arbitraries/valueRanges/OfPrecision.java
jlink: It’s indeed going in the direction I meant, but seeing it it’s unsymmetrical to the Between objects. :-/
So, for symmetry reasons I‘d go back to the previous approach. 

Bringing the logic into OfPrecision, however, should still be an improvement. 

All that forth and back is an implication of asynchronous pull requests. If we were pair-programming, a lot of design ideas could be tried out and would prove themselves worthwhile or unworthy immediately.

zinki97: Isn't it better to change the other classes, too?

jlink:
To me they feel like state holders not like state itself, but I'm unsure.
Which way will make the code simpler in your opinion?

Am Do., 29. Apr. 2021 um 19:06 Uhr schrieb zinki97 ***@***.***

zinki97:
> To me they feel like state holders not like state itself, but I'm unsure. Which way will make the code simpler in your opinion?  .

The previous version is less code, the actual version of `OfPrecision` is more like `LocalDateTime`

jlink:
I vorw for previous version in this case. Until we have new insights.

Am Do., 29. Apr. 2021 um 20:45 Uhr schrieb zinki97 ***@***.***

time/src/main/java/net/jqwik/time/internal/properties/arbitraries/valueRanges/OfPrecision.java
```diff
@@ -34,4 +35,57 @@ public boolean isSet() {
...
+			case MILLIS:
+				return LocalTime.of(23, 59, 59, 999_000_000);
+			case MICROS:
+				return LocalTime.of(23, 59, 59, 999_999_000);
+			case NANOS:
+				return LocalTime.of(23, 59, 59, 999_999_999);
+			default:
+				return LocalTime.of(23, 59, 59, 0);
+		}
+	}
```
zinki97, line 89: I don't know if it is possible to make it simpler without switch, but I think not.

Is the class now like you want or what should I do next?

jlink: Looks good to me.

zinki97: Should I change it in all classes?

Example: `calculateEffectiveMin()`

jlink: Unless you see a reason why calculate* is the better name in some cases

zinki97:
Ready for review

time/src/main/java/net/jqwik/time/internal/properties/arbitraries/DefaultDurationArbitrary.java
```diff
@@ -18,126 +18,73 @@
...
-		} catch (ArithmeticException e) {
-			throw new IllegalArgumentException("Min value must be increased because of precision " + precision + " but results in a " + e.getMessage());
+	private Duration effectiveMin(DurationBetween durationBetween, OfPrecision ofPrecision) {
+		Duration effective = durationBetween.getMin() != null ? durationBetween.getMin() : minPossibleValue(ofPrecision);
+		checkValueAndPrecision(effective, ofPrecision, true);
+		return effective;
+	}
+
+	private Duration minPossibleValue(OfPrecision ofPrecision) {
+		switch (ofPrecision.get()) {
```
jlink, line 54: This calculation should also move to OfPrecision

time/src/main/java/net/jqwik/time/internal/properties/arbitraries/DefaultLocalTimeArbitrary.java
```diff
@@ -51,7 +48,7 @@
 
 	}
 
-	static private LocalTime calculateLocalTime(long longAdd, LocalTime effectiveMin, OfPrecision precision) {
+	static private LocalTime localTimeFromValue(long longAdd, LocalTime effectiveMin, OfPrecision precision) {
 		switch (precision.get()) {
```
jlink, line 52: move calculation to OfPrecision

time/src/main/java/net/jqwik/time/internal/properties/arbitraries/DefaultLocalTimeArbitrary.java
```diff
@@ -68,7 +65,7 @@ static private LocalTime calculateLocalTime(long longAdd, LocalTime effectiveMin
 		}
 	}
 
-	private long calculateLongEnd(LocalTime effectiveMin, LocalTime effectiveMax) {
+	private long longsBetween(LocalTime effectiveMin, LocalTime effectiveMax) {
 		switch (ofPrecision.get()) {
```
jlink, line 69: Move calculation to OfPrecision

time/src/main/java/net/jqwik/time/internal/properties/arbitraries/DefaultLocalTimeArbitrary.java
```diff
@@ -85,97 +82,113 @@ private long calculateLongEnd(LocalTime effectiveMin, LocalTime effectiveMax) {
...
 
-	public static LocalTime calculateEffectiveMaxWithPrecision(LocalTime effective, OfPrecision ofPrecision) {
-		return calculateEffectiveMaxWithPrecision(effective, ofPrecision.get());
-	}
-
-	public static LocalTime calculateEffectiveMaxWithPrecision(LocalTime effective, ChronoUnit ofPrecision) {
-		switch (ofPrecision) {
-			case HOURS:
-				effective = effective.withMinute(0);
+	private static LocalTime setEffectiveMaxValues(LocalTime effective, OfPrecision ofPrecision, ChronoUnit precision) {
```
jlink, line 167: Does the method name fit? Is there something being set?

time/src/main/java/net/jqwik/time/internal/properties/arbitraries/DefaultLocalTimeArbitrary.java
```diff
@@ -85,97 +82,113 @@ private long calculateLongEnd(LocalTime effectiveMin, LocalTime effectiveMax) {
...
+				effective = ofPrecision.get().compareTo(MINUTES) <= 0 ? effective.withMinute(59) : effective.withMinute(0);
 			case SECONDS:
-				effective = effective.withNano(effective.getNano() % 1_000_000);
-			case MILLIS:
-				effective = effective.withNano((effective.getNano() / 1_000_000) * 1_000_000 + effective.getNano() % 1_000);
-			case MICROS:
-				effective = effective.withNano(effective.getNano() - effective.getNano() % 1_000);
+				effective = ofPrecision.get().compareTo(SECONDS) <= 0 ? effective.withSecond(59) : effective.withSecond(0);
+			default:
+				switch (ofPrecision.get()) {
```
jlink, line 174: Move calculation to OfPrecision

time/src/main/java/net/jqwik/time/internal/properties/arbitraries/DefaultLocalTimeArbitrary.java
```diff
@@ -85,97 +82,113 @@ private long calculateLongEnd(LocalTime effectiveMin, LocalTime effectiveMax) {
...
 			case MINUTES:
-				effective = effective.withSecond(0);
+				effective = ofPrecision.get().compareTo(MINUTES) <= 0 ? effective.withMinute(59) : effective.withMinute(0);
 			case SECONDS:
-				effective = effective.withNano(effective.getNano() % 1_000_000);
-			case MILLIS:
-				effective = effective.withNano((effective.getNano() / 1_000_000) * 1_000_000 + effective.getNano() % 1_000);
-			case MICROS:
-				effective = effective.withNano(effective.getNano() - effective.getNano() % 1_000);
+				effective = ofPrecision.get().compareTo(SECONDS) <= 0 ? effective.withSecond(59) : effective.withSecond(0);
```
jlink, line 172: Can those comparisons be replaced by methods in OfPrecision? e.g. ofPrecision.isLessOrEqualTo(..)?

jlink (commented):
I still see potential in giving OfPrecision more functionality that depends on it.
I marked some of those places but certainly not all.
Fix what you see. I'll do the rest after merging the PR. Well, maybe, some day.

jlink (approved):
I think we're done here :-)

PR #193, opened 2021-05-31
add docs to say "Users should use `net.jqwik.api.Tag` rather than `org.junit.jupiter.api.Tag`"

chia7712:
## Overview

The junit 5 `org.junit.jupiter.api.Tag` does not work for jqwik and the replacement is `net.jqwik.api.Tag`. It seems to me that it is worth adding a doc for annotation interpretation.

### Details

I file this PR for (https://github.com/apache/kafka/pull/10791). Kafka leverages jqwik to write tests for new module and I just noticed that junit 5 tag is not compatible to jqwik.

---

### DIFF
--- a/docs/docs/snapshot/user-guide.md
+++ b/docs/docs/snapshot/user-guide.md
@@ -1015,7 +1015,8 @@ class TaggingExamples {
 ```
 
 Tags must follow certain rules as described
-[here](/docs/snapshot/javadoc/net/jqwik/api/Tag.html)
+[here](/docs/snapshot/javadoc/net/jqwik/api/Tag.html). Noted that the `@Tag` you'll
+have to use with jqwik is `net.jqwik.api.Tag` rather than `org.junit.jupiter.api.Tag`
 
 ### Disabling Tests
 


jlink:
@chia7712 Where in the userguide/javadoc would you expect this clarification?

chia7712:
@jlink thanks for your response.

>  Where in the userguide/javadoc would you expect this clarification?

The `Tagging Tests` section in [User Guide](https://jqwik.net/docs/current/user-guide.html#tagging-tests)

docs/docs/snapshot/user-guide.md
```diff
@@ -1015,7 +1015,8 @@ class TaggingExamples {
 ```
 
 Tags must follow certain rules as described
-[here](/docs/snapshot/javadoc/net/jqwik/api/Tag.html)
+[here](/docs/snapshot/javadoc/net/jqwik/api/Tag.html). Noted that the `@Tag` you'll
```
jlink, line 1018: "Noted that" -> "Note that"

jlink (commented):
Sorry for my delayed feedback. There's this minor typo. If you don't find the time I can also fix it myself.

chia7712:
@jlink thanks for approving this patch :)

Issue #201, opened 2021-06-29
Submodules engine, time and web all have the same Module Name


duponter:
## Testing Problem

When adopting the Java Platform Module System in your application, you explicitly list which modules are required to build your application successfully.

When using jqwik's standard features, it is sufficient to require only the _**net.jqwik.api**_ module which is provided by the **_net.jqwik:jqwik-api_** artefact. 

However when you need features from one of jqwik's submodules (time, web or engine), you have to require _**net.jqwik.engine**_ module. However, this displays an _Ambiguous Module Reference_ error as the module is provided by all 3 submodules instead of only one. This makes the functionality of the submodules not usable in combination with the Java Platform Module System.

## Suggested Solution

In the build.gradle files of the mentioned submodules, the jar task explicitly adds an automatic module name. This is, however, only correct for the _**engine**_ submodule and should be corrected in the other two.

jlink:
Thanks for the catch and the pull request. Will merge tomorrow.

jlink:
Available in 1.5.3-SNAPSHOT

jlink:
Released in 1.5.3

PR #202, opened 2021-06-29
Unique Automatic-Module-Name is generated for each submodule jar

duponter:
### Details

Fix for #201 

---


jlink:
How urgent do you need this fix released?

duponter:
It is not blocking for now. 
However, it would be handy if this fix is available in the next couple of weeks, as we will then rely more on the functionality provided by the time submodule.
In any case, thanks for the fast response.

PR #207, opened 2021-07-16
use a data-structure more appropriate for frequently retrieving stores

tmohme:
## Overview

This ist _not_ a ready-to-use fix for the issue - it solves the problem partially and has yet to be verified in a larger context.
At least for my most pathological test class I'm back down to ~10s from 1m45s (1.5.2: ~6s).
So it might be nevertheless good enough for an inspiration.

### Details

There are still some things I find suspicious though:
* I don't expect that many calls to `StoreRepository#get` (relative to the number of tries)
* I don't expect a (slow but) steady increase in entries for a given test during execution of the tries
* The test execution time increases by 20x when I increase the tries by 10x (1.000 -> 10.000)
* The implementation has still room for improvement for the other public methods

---

### DIFF
--- a/engine/src/main/java/net/jqwik/engine/execution/lifecycle/StoreRepository.java
+++ b/engine/src/main/java/net/jqwik/engine/execution/lifecycle/StoreRepository.java
@@ -25,7 +25,9 @@ public class StoreRepository {
 		return current;
 	}
 
-	private final Set<ScopedStore<?>> stores = new HashSet<>();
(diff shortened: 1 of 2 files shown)


jlink:
@tmohme Lots of thanks. I think I'll first track down why getStore is called so often and then pull in your implementation of IdentifiedStores.

Do you have a smallish property where the acceleration is measurable?

jlink:
I copied the code over by hand since draft PRs cannot be merged.

PR #211, opened 2021-07-31
Bug Fixes and Instant

zinki97:

- Bug fixes
- arbitrary for type `Instant`

### DIFF
--- a/time/src/main/java/net/jqwik/time/api/arbitraries/InstantArbitrary.java
+++ b/time/src/main/java/net/jqwik/time/api/arbitraries/InstantArbitrary.java
@@ -1,0 +1,19 @@
+package net.jqwik.time.api.arbitraries;
+
+import java.time.*;
+
+import org.apiguardian.api.*;
+
+import net.jqwik.api.*;
+
+import static org.apiguardian.api.API.Status.*;
+
+/**
+ * Fluent interface to configure the generation of instant values.
+ * All generated instants use the Gregorian Calendar, even if they are before October 15, 1582.
+ * By default, instants with years between 1900 and 2500 are generated.
+ */
+@API(status = EXPERIMENTAL, since = "1.5.5")
+public interface InstantArbitrary extends Arbitrary<Instant> {
+
+}
--- a/time/src/main/java/net/jqwik/time/api/arbitraries/InstantArbitrary.java
+++ b/time/src/main/java/net/jqwik/time/api/arbitraries/InstantArbitrary.java
@@ -15,7 +15,7 @@ import static org.apiguardian.api.API.Status.*;
  * By default, instants with years between 1900 and 2500 are generated.
  * Max possible year is 999_999_999.
  */
-@API(status = EXPERIMENTAL, since = "1.5.5")
+@API(status = EXPERIMENTAL, since = "1.5.4")
 public interface InstantArbitrary extends Arbitrary<Instant> {
 
 	/**
--- a/time/src/main/java/net/jqwik/time/api/arbitraries/InstantArbitrary.java
+++ b/time/src/main/java/net/jqwik/time/api/arbitraries/InstantArbitrary.java
@@ -44,7 +44,6 @@ public interface InstantArbitrary extends Arbitrary<Instant> {
 	/**
 	 * Set the allowed lower {@code min} (included) and upper {@code max} (included) bounder of generated local date values.
 	 */
-	@API(status = EXPERIMENTAL, since = "1.5.2")
 	InstantArbitrary dateBetween(LocalDate min, LocalDate max);
 
 	/**
@@ -54,7 +53,6 @@ public interface InstantArbitrary extends Arbitrary<Instant> {
 	 * <p>Calling this method is equivalent to calling {@linkplain #dateBetween(LocalDate, LocalDate)}
 	 * assuming Jan 1 and Dec 31 as first and last day of those years.</p>
 	 */
-	@API(status = EXPERIMENTAL, since = "1.5.2")
 	InstantArbitrary yearBetween(Year min, Year max);
 
 	/**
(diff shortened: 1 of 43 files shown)


zinki97:
citation from https://github.com/jlink/jqwik/issues/175

Me:
> A few weeks ago you said that the tests in Calender could look like this: Calender behaves likes LocalDateTime.
Can you please give an example how such a test case could look like? @jlink

@jlink :
> Looking over Calendar I guess I was wrong. Calendar instances have a timezone. Moreover, calendar instances can be created from different calendar types. So it might make sens to generate instances of the concrete subclasses Gregorian, Buddhist and JapaneseImperial. Maybe Buddhist and japanese imperial are out of scope.

Maybe I can use such test cases here? So again: Can you please give an example how such a test case could look like? (for Instant)

jlink:
What I meant is that tests for Instant will more or less be the same as for LocalDateTime.

> Edit: I think the time zone for Instant is always converted to "Z".
Exactly.

Or maybe I don't understand your question.

zinki97:
> What I meant is that tests for Instant will more or less be the same as for LocalDateTime.
> 
> > Edit: I think the time zone for Instant is always converted to "Z".
> > Exactly.
> 
> Or maybe I don't understand your question.

I unterstood that I should not just copy the tests, but write tests that test whether the two classes behave the same way. Am I wrong? Otherwise: How can I do that?

jlink:
> I unterstood that I should not just copy the tests, but write tests that test whether the two classes behave the same way. Am I wrong? Otherwise: How can I do that?

That's fairly straightforward when two implementations have a common supertype / iterface that reflects all necessary functionality Instant and LocalDateTime have Temporal, which probably could be used for some tests.
But I don't think that's a worthwhile abstraction in this case.

If generation of Instant will be nothing but generating LocalDateTime instances and mapping them to Instant, you could probably spare most tests that cover special cases and just test that Instant will be generated at all when requested.

If, however, Instant generation requires more than trivial logic, the tests covering this logic will have to be duplicated I'm afraid.

zinki97:
What is currently missing: The documentation

jlink:
Is this PR ready for review?

zinki97:
> Is this PR ready for review?

now

time/src/main/java/net/jqwik/time/internal/properties/arbitraries/DefaultInstantArbitrary.java
```diff
@@ -0,0 +1,128 @@
...
+
+import static org.apiguardian.api.API.Status.*;
+
+@API(status = INTERNAL)
+public class DefaultInstantArbitrary extends ArbitraryDecorator<Instant> implements InstantArbitrary {
+
+	private LocalDateTimeArbitrary dateTimeArbitrary;
+
+	public DefaultInstantArbitrary() {
+		dateTimeArbitrary = DateTimes.dateTimes();
```
jlink, line 21: Why not assign in member variable declaration?

time/src/main/java/net/jqwik/time/internal/properties/arbitraries/DefaultInstantArbitrary.java
```diff
@@ -0,0 +1,128 @@
...
+
+	@Override
+	protected Arbitrary<Instant> arbitrary() {
+
+		return dateTimeArbitrary.map(dateTime -> dateTime.toInstant(ZoneOffset.UTC));
+
+	}
+
+	public static LocalDateTime instantToLocalDateTime(Instant instant) {
+		if (instant.isAfter(LocalDateTime.MAX.toInstant(ZoneOffset.UTC))) {
```
jlink, line 32: Use Instant.MAX instead?

time/src/main/java/net/jqwik/time/api/arbitraries/InstantArbitrary.java
```diff
@@ -0,0 +1,128 @@
...
+
+import static org.apiguardian.api.API.Status.*;
+
+/**
+ * Fluent interface to configure the generation of instant values.
+ * All generated instants use the Gregorian Calendar, even if they are before October 15, 1582.
+ * By default, instants with years between 1900 and 2500 are generated.
+ * Max possible year is 999_999_999.
+ */
+@API(status = EXPERIMENTAL, since = "1.5.5")
```
jlink, line 18: Next version is 1.5.4. Should be doable is you don't plan for more functionality in this PR.

time/src/main/java/net/jqwik/time/api/constraints/InstantRange.java
```diff
@@ -0,0 +1,24 @@
+package net.jqwik.time.api.constraints;
+
+import java.lang.annotation.*;
+
+import org.apiguardian.api.*;
+
+import static org.apiguardian.api.API.Status.*;
+
+/**
+ * Constrain the range of generated date time parameters.
```
jlink, line 10: instant

time/src/main/java/net/jqwik/time/internal/properties/arbitraries/DefaultInstantArbitrary.java
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	@Override
+	protected Arbitrary<Instant> arbitrary() {
+
+		return dateTimeArbitrary.map(dateTime -> dateTime.toInstant(ZoneOffset.UTC));
+
+	}
+
+	public static LocalDateTime instantToLocalDateTime(Instant instant) {
```
jlink, line 31: Is it public only to support tests? Maybe duplicate it somewhere in the test packages.

jlink (commented):
Looks good.

time/src/main/java/net/jqwik/time/internal/properties/arbitraries/DefaultInstantArbitrary.java
zinki97: No it's also in some configurators.

time/src/main/java/net/jqwik/time/internal/properties/arbitraries/DefaultInstantArbitrary.java
zinki97: This is not possible, because `Instant.MAX` has the year `1,000,000,000`. I decided to exclude this year so that we can use the `LocalDateArbitrary`.

PR #212, opened 2021-08-11
OffsetDateTimes

zinki97:


jlink:
Can you add `@StatisticsReport(onFailureOnly = true)` to all `EqualDistributionTests` groups and other container classes that use statistics checks? This will take care that statistics only show up in the report if the property fails.

zinki97:
> Can you add `@StatisticsReport(onFailureOnly = true)` to all `EqualDistributionTests` groups and other container classes that use statistics checks? This will take care that statistics only show up in the report if the property fails.

Done. I don't know if I've found all statistic tests. Do you see something missing?

jlink:
You can run the tests and check if something is logged.

jlink:
The only remaining logging I can see is:
```
InvalidCombinationTests > hourMinuteSecondBetweenAndBetween STANDARD_ERROR
    Aug 24, 2021 3:33:23 PM net.jqwik.engine.properties.RandomizedShrinkablesGenerator logEdgecasesOutnumberTriesIfApplicable
    INFO: Edge case generation exceeds number of tries. Stopped after 6000 generated cases.

InvalidCombinationTests > hourMinuteSecondTimeBetween STANDARD_ERROR
    Aug 24, 2021 3:33:24 PM net.jqwik.engine.properties.RandomizedShrinkablesGenerator logEdgecasesOutnumberTriesIfApplicable
    INFO: Edge case generation exceeds number of tries. Stopped after 5000 generated cases.

```
Annotating `InvalidCombinationTests` with `@SuppressLogging` should do the trick.

zinki97:
> The only remaining logging I can see is:

Thank you, I've overlooked this.

PR #213, opened 2021-08-16
Fix link into vavr-test

valery1707:
## Overview

Module `vavr-test` was moved out from `vavr` in vavr-io/vavr-test@8523bf55f12ade9de3f0179a0724910c125063b6.

---


jlink:
Thanks for the catch!

PR #234, opened 2021-10-04
Change Memoize cache to LruCache

mhyeon-lee:
## Overview

Change to LruCache so that data is not continuously accumulated in the static Store HashMap of Memoize .

### Details

Memoize #memodizedGenerator caches RandomGenerator to Memoize HashMap as many as the number of Arbitrary types (`Tuple3<Arbitrary<?>, Integer, Boolean>`).
When many kinds of Arbitrary are created, Random Generators accumulate in Map.
Change to prevent accumulation of too many Random Generators with LRU Cache.

---


jlink:
Deployed to 1.6.0-SNAPSHOT

PR #235, opened 2021-10-04
Enhance SampleStreamFacade runInDescriptor empty case finish own scope

mhyeon-lee:
## Overview

Enhance SampleStreamFacade runInDescriptor empty case finish own scope.

### Details

`CurrentTestDescriptor.isEmpty()` is not common case.
Depending on how SampleStream is called, the case may occur.
ex: run sampleStream outside the jqwik thread or class static scope.

Use the temporary Descriptor and then finish it within its own scope.
To prevent leakage of the store

---

### DIFF
--- a/engine/src/main/java/net/jqwik/engine/facades/SampleStreamFacade.java
+++ b/engine/src/main/java/net/jqwik/engine/facades/SampleStreamFacade.java
@@ -56,7 +56,15 @@ class SampleStreamFacade {
 
 	private static <T> T runInDescriptor(Supplier<T> code) {
 		if (CurrentTestDescriptor.isEmpty()) {
-			return CurrentTestDescriptor.runWithDescriptor(SAMPLE_STREAM_DESCRIPTOR, code);
+			// CurrentTestDescriptor.isEmpty is not common case.
+			// ex: run sampleStream outside the jqwik thread or class static scope.
+			// Use the temporary Descriptor and then finish it within its own scope.
+			// (To prevent leakage of the store)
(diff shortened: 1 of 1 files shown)


jlink:
I don't think your suggested solution works is all cases. 

Some generators require access to a shared store. With your solution they'd not share the same store since it would be recreated for each sample. Memoization is one of those cases, i.e. with your suggested change caching of generators would not work outside the jqwik lifecycle. 

To see that try to debug this simple JUnit 5 test:
```
	@Test
	void samplingOutsideJqwik() {
		Arbitrary<String> strings = Arbitraries.strings().alpha().ofLength(5);
		strings.sampleStream().limit(10).forEach(v -> System.out.println(v));
	}
```
with a breakpoint in line 31 of Memoize.java

A working solution would somehow make sure that the lifecycle for descriptor SAMPLE_STREAM_DESCRIPTOR is being executed within

```
class SampleStreamFacade...
	<T> Stream<T> sampleStream(Arbitrary<T> arbitrary) {
		RandomGenerator<T> generator = getGeneratorForSampling(arbitrary);
		return Stream.generate(wrapInDescriptor(() -> generator.next(SourceOfRandomness.current())))
					 .map(shrinkable -> runInDescriptor(() -> shrinkable.value()));
	}
```

This might require a dedicated Stream implementation to call `finishScope` only when stream evaluation has stopped.

mhyeon-lee:
@jlink 
Thank you for your advice. 
I agree with the opinion that it is an inappropriate solution.

I will close this PR and raise the result being tested as an issue.

PR #253, opened 2021-11-13
Configure version to be included in module descriptors

sormuras:
Due to an update of the default behaviour of beryx's module plugin to no longer include project's version as module version, this PR adds explicit configurations to achieve that.

---

### DIFF
--- a/api/build.gradle
+++ b/api/build.gradle
@@ -7,6 +7,7 @@ jar {
 
 moduleConfig {
 	moduleInfoPath = 'src/main/module/module-info.java' // don't confuse IDEs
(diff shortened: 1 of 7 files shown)


sormuras:
@beryx / @siordache what does this line configure exactly?

```
neverCompileModuleInfo = true
```

jlink:
I wonder if the moduleInfo block could go into the parent Gradle file?

sormuras:
> I wonder if the moduleInfo block could go into the parent Gradle file?

True. Each block looks very similar to each other...

PR #259, opened 2021-11-21
Documentation: Fix name of `jqwik-kotlin`

goerge:
The documentation on https://jqwik.net/docs/current/user-guide.html#kotlin-module reads:

> This module’s artefact name is jqwik-Module.

where it should be

> This module’s artefact name is jqwik-kotlin. 

This PR fixes this.

---


jlink:
@goerge Thanks for the catch

Issue #108, opened 2020-05-28
Provide user guide as PDF


jlink:
It would be useful to have the user-guide as a downloadable PDF file.

PDF should be linked on homepage.

metters:
Does it mean, the PDF is generated when building the project with gradle? 
The link to the homepage is https://jqwik.net/?

jlink:
The idea is to generate it after generating the documentation, which is a task of its own in https://github.com/jlink/jqwik/blob/master/documentation/build.gradle. It would probably require to include some  of the markdown to pdf tools.

Bluesoul02:
Would it be ok to generate the pdf using pandoc just like in the pull request?

jlink:
@Bluesoul02 I’ll have a look at it. Tomorrow.

jlink:
@Bluesoul02 Many thanks for tackling the issue! Let's work it out in the PR.

PR #271, opened 2021-12-06
WIP - Generate a pdf of the user guide

Bluesoul02:
Generate a pdf of the user guide using pandoc

## Overview

#108 

### Details

pandoc and xelatex need to be installed to generate the documentation

the pdf is not looking good for now

---

### DIFF
--- a/documentation/build.gradle
+++ b/documentation/build.gradle
@@ -98,8 +98,16 @@ task expandJavaDocIndexHtml(type: Copy) {
 	)
 }
 
+task generateUserGuidePDF(type : Exec) {
+	dependsOn generateUserGuideToc
+
+	workingDir "$buildDir/docs"
+
+	commandLine 'pandoc', '--pdf-engine=xelatex', '-s', '-o', 'user-guide.pdf', 'user-guide.md'
+}
+
 task generateDocumentation {
-	dependsOn(generateUserGuideToc, generateJavaDoc, expandJavaDocIndexHtml)
+	dependsOn(generateUserGuideToc, generateJavaDoc, expandJavaDocIndexHtml, generateUserGuidePDF)
 	doLast {
 		String docsFolder = "${buildDir}/docs-${docsVersion(jqwikVersion)}"
 		file(docsFolder).delete()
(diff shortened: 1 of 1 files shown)


jlink:
I installed pandoc through homebrew and tried to generated the documentation. It failed.

So I cded to `documentation/build/docs`, which already had `user-guide.md` and executed 
```
pandoc --pdf-engine=xelatex -s -o user-guide.pdf user-guide.md 
```
Got the following error:
`pandoc: xelatex: createProcess: posix_spawnp: illegal operation (Inappropriate ioctl for device)`

Any idea what might be going on?

Bluesoul02:
It might be an error specific to Mac OS : https://github.com/jgm/pandoc/issues/7570 
it looks like it's a problem with xelatex installation, adding xelatex to the PATH might solve the problem, we'll try to see if we can run it on Mac OS on our side.

jlink:
> It might be an error specific to Mac OS : [jgm/pandoc#7570](https://github.com/jgm/pandoc/issues/7570) it looks like it's a problem with xelatex installation, adding xelatex to the PATH might solve the problem, we'll try to see if we can run it on Mac OS on our side.

I had to install another package that contains xelatex. On mac that's `mactex` which is actually very large and took half an hour to install.

Do you think there's a way with a smaller footprint for generating a pdf?

Bluesoul02:
So I found two alternatives that have a smaller footprint, but I'm still working on getting them to work.  
https://github.com/fntsoftware/gradle-plugin-markdown2pdf  
https://pypi.org/project/Markdown2PDF/

documentation/build.gradle
```diff
@@ -98,8 +98,16 @@ task expandJavaDocIndexHtml(type: Copy) {
 	)
 }
 
+task generateUserGuidePDF(type : Exec) {
+	dependsOn generateUserGuideToc
+
+	workingDir "$buildDir/docs"
+
+	commandLine 'pandoc', '--pdf-engine=xelatex', '-s', '-o', 'user-guide.pdf', 'user-guide.md'
```
vlsi, line 106: Should this specify `task inputs` and `task outputs`?
Inputs and outputs would enable Gradle to track up-to-date properly, and the task could even become build-cacheable in the future.

PR #282, opened 2021-12-31
introduce 'jqwik.common-configuration.gradle' to share build logic

tmohme:
## Overview

Refactor Gradle build structure towards using a convention plugin instead of cross project configuration which has a negative impact on coupling and prevents optimizations (by Gradle during an actual build).

### Details

Because Gradle doesn't allow arbitrary code before 'build script {}‘ / 'plugins {}', it was required to extract the extra properties to an extra file ("dependencies.gradle").
They dependency versions _could_ also be placed either in their own file under 'buildSrc' (worked well in other projects), integrated in the convention plugin (imho a bit "buried") or specified with Gradle's new "version catalog" feature.
Unfortunately, none of those alternatives is currently understood by Dependabot.

---


jlink (commented):
I think that's really cool. Could the following block also move to the common-configuration plugin?

moduleConfig {
	moduleInfoPath = 'src/main/module/module-info.java' // don't confuse IDEs
	version = "${jqwikVersion}"
	neverCompileModuleInfo = true
}

And would you rebase once more since a few versions have changed meanwhile?

tmohme:
I saw that not all Gradle subprojects actually have a `moduleConfig`, therefore I think it would be cleaner to create another "jqwik.module-configuration.gradle"  plugin, containing just the application and configuration of the "org.beryx.jar" plugin and apply it only to the relevant subprojects which create jar artifacts. 
What do you think?

tmohme:
@jlink introduced 'jqwik.module-configuration.gradle' to share build logic & rebased

jlink:
@tmohme Many thanks!

PR #287, opened 2022-01-06
fix typos in documentation

vlsi:
Hi Johannes,

I'm planning to implement InputStream tests for https://github.com/pgjdbc/pgjdbc/pull/2389, and I discovered a couple of typos in the documentation.

The doc is great: it does list up-to-date jqwik version, the samples look good.

What I plan to implement is a mix of `Stateful Testing` and `Contract Tests`.
In other words, I have a golden implementation (e.g. `ByteArrayInputStream`), and I would verify if my `InputStream` yields the same (or comparable) outputs where the operations are `read(), read(byte[]), read(byte[], int, int), mark(int), reset()`.

It would be great to decouple `InputStreamProperty` from `pgjdbc`, so `InputStreamProperty` could be contributed back to jqwick.

---

...migrated to https://github.com/jlink/jqwik/issues/288

---

### DIFF
--- a/documentation/src/docs/include/generation-from-type.md
+++ b/documentation/src/docs/include/generation-from-type.md
@@ -32,7 +32,7 @@ public class Person {
 ```
 
 A first step to use arbitrarily generated `Person` objects without having
-to write a lot of _jqwik_-specific boiler plat code could look like that:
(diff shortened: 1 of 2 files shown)


jlink:
@vlsi Many thanks!

Issue #250, opened 2021-11-12
More Kotlin Support


jlink:
## Testing Problem

With version 1.6.0 jqwik offers an optional kotlin module.
Not everything from the initial issue https://github.com/jlink/jqwik/issues/151 has been implemented yet.

## Suggested Solution

Add the following Kotlin-related features:
- Support nullability detection in nested types like `@ForAll list: List<String?>`.
  This will require that jqwik uses Kotlin reflection for creating the `TypeUsage` values for parameters.
  Currently only Java reflection is used with a bit of enhancement through `TypeUsage.Enhancer`
  - Having that in-place correct handling of unsigned integer types should be possible
- More generators and default generation:
  - CharRange
  - LongRange
  - DoubleRange 
  - Progressions
- Support kotlin.contracts
- Support generation of types from common K libs:
  - https://arrow-kt.io/

vlsi:
One more item: declaration-site variance.

Currently, Kotlin treats Java's `Arbitrary<T>` as invariant.
In other words, if I declare `interface ArbitraryAction<T> : Arbitrary<MyActionType<T>>` (e.g. to reduce nesting everywhere), then Kotlin does not allow returning sub-types without casts.

----

The idea of Kotlin generics is that you can specify "usage type" once at type parameter declaration, and then it would be reused in all the usages.

In other words, in Kotlin, the following would mean "the interface only produces objects of `T`, and it never consumes them":

```kotlin
interface Arbitrary<out T> {
...
```

Sample:

```kotlin
interface RegularArbitrary<T>
interface OutArbitrary<out T>

class Test {
    lateinit var regularString: RegularArbitrary<String>
    lateinit var regularAny: RegularArbitrary<Any>

    lateinit var outString: OutArbitrary<String>
    lateinit var outAny: OutArbitrary<Any>

    fun withRegularArbitrary(regularAny: RegularArbitrary<Any>) {
    }
    fun withOutArbitrary(regularAny: OutArbitrary<Any>) {
    }

    fun test() {
        // Type mismatch. Required: RegularArbitrary<Any> Found: RegularArbitrary<String>
        withRegularArbitrary(regularString)
        // Compiles ok
        withOutArbitrary(outString)
    }
}
```

In the ideal world, there should be an annotation (or something else) that could clarify Kotlin compiler that given interface declaration `Arbitrary<T>` means `Arbitrary<out T>`.

[...]

PR #302, opened 2022-02-09
WIP: migrate API interfaces to Kotlin for adding generic variance

vlsi:
## Overview

See sample in https://github.com/jlink/jqwik/issues/250#issuecomment-1031412461

It is common wisdom that "producer extends consumer super", so it would be better to use `Function<? super T, ? extends R>` instead of `Function<T, R>`; `Predicate<? super T>`

See https://stackoverflow.com/questions/2723397/what-is-pecs-producer-extends-consumer-super

It might be valid to use `Option<? extends T>` in method return values (see clarifications in https://github.com/SpongePowered/SpongeAPI/issues/369 , https://github.com/SpongePowered/SpongeAPI/pull/386 )

---

Kotlin surfaces the issues since its generic variance is declared at the class/interface declaration, and then it requires that type parameter should be used with the same variance (there are annotations `@JvmSuppressWildcards` and `@JvmWildcard` to augment the behavior)

For instance:

```kotlin
interface Arbitrary<out T> { // <-- the interface "produces T", which is more or less the same as writing `? extends T` everywhere in Java

    // "in T" is required otherwise the error is
    // Type parameter T is declared as 'out' but occurs in 'invariant' position in type Function<T, U>
    // out U is to be added manually
    fun <U> map(mapper: Function<in T, out U>): Arbitrary<U> {
    // translates to:
    // <U> Arbitrary<U> map(Function<? super T, ? extends U>);

    // Predicate<List<T>> is a compilation error in Kotlin
    // Predicate<List<T>>
    fun collect(until: Predicate<in List<T>>): Arbitrary<List<T>> {
    // translates to:
    // Arbitrary<java.util.List<T>> collect(Predicate<? super List<? extends T>>)
```

### Details

Note: this is just a prototype to see what the implementation might look like.

### Findings

1. kotlin-stdlib seems to be needed for compilation. I'm not sure there's a way to generate `int` bytecode, and so on.

For instance:
```
Cannot access built-in declaration 'kotlin.Int'
Cannot access built-in declaration 'kotlin.Throwable'
Unresolved reference: JvmStatic
Cannot access built-in declaration 'kotlin.Comparable'
```

2. The change surfaced a lot `Function<T, R>`, `Predicate<T>`. I believe it makes sense to add `extends` and `super` wildcards even without Kotlin.

3. There are methods that return `Optional<T>` . Kotlin wants them to be `Optional<out T>` (it is in line with https://github.com/SpongePowered/SpongeAPI/issues/369 ). However, it makes Java usages a bit more verbose (see sample diffs in test classes like `Arbitrary<List<? extends String>> addressParts = ipv6Part().list().ofSize(8)`

4. The compilation time seems to be intact: `:api` is very small, so it takes just a couple of seconds to compile it.


jlink:
@vlsi thanks for putting time into it. I don't think, though, I'll make jqwik's core dependent on Kotlin in any way anytime soon; even if it's only at build time. My experience (my guts) tell me that steering clear of dependencies is the higher good here.

As for the underlying problem you are facing, I'll look at other approaches to at least mitigate the nuisance.

PR #344, opened 2022-05-26
Use Arrays.binarySearch in ChooseRandomlyByFrequency

### LINKED ISSUE #332: Use Collections/Arrays#binarySearch in ChooseRandomlyByFrequency
## Problem

The current [ChooseRandomlyByFrequency](https://github.com/jlink/jqwik/blob/main/engine/src/main/java/net/jqwik/engine/support/ChooseRandomlyByFrequency.java) uses a brute force approach when selecting the corresponding element.

## Suggested Solution

Use `List<Tuple.Tuple2<Integer, T>>` or `int[] weights + Object[] values` instead of `Map<T, Integer> upperBorders`.

## Discussion

The key driver for me is code readability rather than performance. I ran into `ChooseRandomlyByFrequency` in the debugger.

Having linear structure + binary search should be faster than querying `HashMap.get` multiple times.
At the end of the day, we can even use `int[] weights` + `Object[] values` storage, so both `Arrays.binarySearch` and `sequential scan over weights` are possible.

## Known algorithms

* Alias method: https://en.wikipedia.org/wiki/Alias_method

> The algorithms typically use O(n log n) or O(n) preprocessing time, after which random values can be drawn from the distribution in O(1) time

* The Fast Loaded Dice Roller: https://github.com/probcomp/fast-loaded-dice-roller

The authors claim FLDR is faster than Alias.

https://arxiv.org/pdf/2003.03830v2.pdf

vlsi:
## Overview

`ChooseRandomlyByFrequency#choose` is slow as it scans elements one by one.

### Details

fixes #332

---

### DIFF
--- a/engine/src/main/java/net/jqwik/engine/support/ChooseRandomlyByFrequency.java
+++ b/engine/src/main/java/net/jqwik/engine/support/ChooseRandomlyByFrequency.java
@@ -7,7 +7,7 @@ import net.jqwik.api.*;
 
 public class ChooseRandomlyByFrequency<T> implements Function<Random, T> {
 
-	private final Map<T, Integer> upperBorders = new HashMap<>();
+	private int[] upperBounds;
 	private int size = 0;
 	private List<T> valuesToChooseFrom;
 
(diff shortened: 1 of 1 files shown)


engine/src/main/java/net/jqwik/engine/support/ChooseRandomlyByFrequency.java
```diff
@@ -26,32 +26,39 @@ protected List<T> possibleValues() {
...
+		} else {
+			// Exact value is found, use the next bucket
+			// For instance, if weights are {2,3}, the upperBounds are {2,5},
+			// and the input indices could be 0..4.
+			// "index" {0, 1} should be mapped to the first element
+			// "index" {2, 3, 4} should be mapped to the second
+			// If (input) index==2, binarySearch returns 0, so we must advance i by one
+			// This advance is never triggered for the very last element since random.nextInt
+			// excludes upper bound.
+			i++;
```
vlsi, line 59: This logic won't be required if we code the binary search directly instead of calling to `Arrays.binarySearch`, however, I think it is more readable to use `Arrays.binarySearch`.

jlink:
Thanks @vlsi!

PR #345, opened 2022-05-26
Avoid recursion in CombinedIterator

### LINKED ISSUE #342: Avoid recursion in CombinedIterator
## Testing Problem

I need to create database tables, and I want to test extreme cases like "1000 columns in a single table".

My code looks as follows:

```kotlin
// Generate a valid column name: [A-Z][A-Z0-9#$_]{0,29}
val arbitraryColumnName = combine(
    Arbitraries.chars().range('A', 'Z'),
    Arbitraries.strings().withCharRange('A', 'Z').withCharRange('0', '9')
        .withChars('#', '$', '_')
        .ofMinLength(0).ofMaxLength(29)
) { first, rest -> first + rest }

context(MappingState)
fun `add single table`() =
    `add table base`(
        // I need up to 1000 columns, each should have a different name
        arbitraryColumnName.set().ofMinSize(1).ofMaxSize(1000).flatMapEach { _, u ->
            `add column`(just(u), ...)
        }
    )
```

The current implementation causes extremely long stacktraces, and it is slow for cases like 1000 elements (~3 seconds for fetching 1000 elements out of CombinedIterator).
One of the pitfalls is that each of `net.jqwik.engine.support.combinatorics.CombinedIterator#next` creates its own `List`, so a top-level `next` call causes O(N^2) array allocations.

## Suggested Solution

Use `Iterator[]` or `List<Iterator>` in `CombinedIterator` and draw from the iterators as needed.

vlsi:
Previously, CombinedIterator#next copied the list as many times
as there were elements in the list (each recursive call did copy the list).

Now only a single copy is made.

It improves performance, and it makes stacktraces easier to understand.

fixes #342

---

### DIFF
--- a/engine/src/main/java/net/jqwik/engine/support/combinatorics/CombinedIterator.java
+++ b/engine/src/main/java/net/jqwik/engine/support/combinatorics/CombinedIterator.java
@@ -54,8 +54,14 @@ public class CombinedIterator<T> implements Iterator<List<T>> {
 	}
 
 	private void resetValuesFrom(int startPosition) {
+		List<Iterator<T>> iterators = this.iterators;
+		List<Iterable<T>> iterables = this.iterables;
+		List<T> elements = this.elements;
+		// In the initial reset, we can reuse the existing iterators
+		// It might slightly optimize the behavior, and it supports Stream#iterator which can't be executed twice
+		boolean initialReset = position == -1;
 		for (int i = startPosition; i < iterables.size(); i++) {
-			Iterator<T> newIt = iterables.get(i).iterator();
+			Iterator<T> newIt = initialReset ? iterators.get(i) : iterables.get(i).iterator();
 			iterators.set(i, newIt);
 			elements.set(i, newIt.next());
 		}
@@ -63,6 +69,7 @@ public class CombinedIterator<T> implements Iterator<List<T>> {
 	}
 
 	private int nextAvailablePosition() {
+		List<Iterator<T>> iterators = this.iterators;
 		for (int i = position; i >= 0; i--) {
 			if (iterators.get(i).hasNext()) {
 				return i;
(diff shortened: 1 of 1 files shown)


vlsi:
This change improves `while (iterator.hasNext() && count < maxEdgeCases) {` loop in `net.jqwik.engine.properties.arbitraries.EdgeCasesSupport#combine` from 2800 ms to something like 50ms when generating 1000 cases out of 1000 arbitraries.

vlsi:
I see the test failed because my implementation prefers shrinking the first.

Previous impl:
`net.jqwik.api.CombinatorsTests.ExhaustiveGeneration#combine3arbitraries`

https://github.com/jlink/jqwik/blob/1312f594adf05a5859871a4a3b1649baf905aff7/engine/src/test/java/net/jqwik/api/CombinatorsTests.java#L320-L331

Expected: `111, 112, 121, 122, 211, 212, 221, 222`
New impl: `111, 211, 121, 221, 112, 212, 122, 222`

Is it crucial to advance the last arbitrary first?
Should I update the test data to the new order?

vlsi:
I added a commit that restores the previous iteration order. Frankly speaking, I think iterating from "last elements to the first" is slightly less readable, however, it might be good to keep the existing iteration order so people have the less disruptions (e.g. the same seeds result in the same results) as they update jqwik versions.

jlink:
@vlsi Thanks!

engine/src/main/java/net/jqwik/engine/support/combinatorics/CombinedIterator.java
```diff
@@ -2,62 +2,79 @@
...
-				   ? emptyListIterator()
-				   : new CombinedIterator<>(this.rest);
-	}
-
-	private Iterator<List<T>> emptyListIterator() {
-		return asList(Collections.<T>emptyList()).iterator();
+		this.iterables = iterables;
+		elements = new ArrayList<>(Collections.nCopies(iterables.size(), null));
+		iterators = iterables.stream().map(Iterable::iterator).collect(Collectors.toCollection(ArrayList::new));
+		isEmpty = !iterators.stream().allMatch(Iterator::hasNext);
```
vlsi, line 18: @jlink , I just noticed `allMatch` would return `true` for the empty stream, so the expression should be adjusted as `isEmpty = iterables.isEmpty() || !iterators.str...`

jlink:
fixed

Issue #281, opened 2021-12-30
Gradle Kotlin DSL build script


tmohme:
## Testing Problem

It does not solve any particular testing problem ;)

## Suggested Solution

If you are interested in having type-safe Kotlin DSL build scripts, I would try to convert the existing ones (plus changing the structure to avoid cross-project configuration) in a PR.

## Discussion

Structurally, the result would look similar to what is described at https://docs.gradle.org/current/samples/sample_building_kotlin_applications_multi_project.html (especially the buildSrc).

jlink:
@tmohme Thanks for the suggestion. I prefer the Groovy syntax of Gradle, but that's just a personal preference given my Groovy history. I have the feeling that it gives me a bit more leeway and is less verbose in some cases.

But what do you mean by "plus changing the structure to avoid cross-project configuration"? That sounds interesting.

tmohme:
In the "early days" of Gradle, using "subprojects {}" and "allprojects {}" were the only means to share build logic between Subprojects.
Since a few major releases there is another, now recommended approach: Using convention plugins.
See https://docs.gradle.org/current/userguide/sharing_build_logic_between_subprojects.html#sec:convention_plugins_vs_cross_configuration
From my experience, I can confirm the described problems with cross project configuration and I had no issues with the convention plugin approach (yet).

jlink:
This make sense - even if the Gradle documentation requires quite some hopping to understand how it works. 

@tmohme If you’re willing to start with convention plugins and keep the Groovy DSL (at least in the beginning) I’d be more than happy.

vlsi:
I recently migrated https://github.com/allure-framework/allure-gradle to the [idiomatic structure](https://github.com/jjohannes/idiomatic-gradle), and I incline to like that structure (it is literally the same as the one used by Gradle itself)

The only bit I would like to do slightly different is to name `build-logic` as `allure-gradle-build-logic` since currently Gradle does not support composite builds where multiple builds have their own `build-logic` module (see https://github.com/gradle/gradle/issues/17228).
The workable workaround is to use `${project_name}-build-logic` or `build-logic-${project_name}` naming.

PR #349, opened 2022-05-28
Add com.github.vlsi.gradle-extensions to make test output human-readable

vlsi:
## Overview

I might be biased, however, I believe https://github.com/vlsi/vlsi-release-plugins/blob/master/plugins/gradle-extensions-plugin/README.md#test-output-formatting makes CI output much easier to understand.

---

### DIFF
--- a/buildSrc/src/main/groovy/jqwik.common-configuration.gradle
+++ b/buildSrc/src/main/groovy/jqwik.common-configuration.gradle
@@ -2,6 +2,8 @@ plugins {
 	id 'java-library'
 	id 'maven-publish'
 	id 'signing'
+	// See https://github.com/vlsi/vlsi-release-plugins/blob/master/plugins/gradle-extensions-plugin/README.md
+	id 'com.github.vlsi.gradle-extensions'
 }
 
 repositories {
--- a/buildSrc/build.gradle
+++ b/buildSrc/build.gradle
@@ -11,3 +11,7 @@ repositories {
     // Use the plugin portal to apply community plugins in convention plugins.
     gradlePluginPortal()
 }
+
+dependencies {
+	implementation 'com.github.vlsi.gradle-extensions:com.github.vlsi.gradle-extensions.gradle.plugin:1.78'
+}
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,0 +1,3 @@
+# com.github.vlsi.gradle-extensions prints only failing or slow test results
+slowTestLogThreshold=15000
+slowSuiteLogThreshold=30000


jlink:
I assume that also works for output in GitHub actions?

vlsi:
Yeah, I mainly implemented the plugin for improving GitHub Actions output.
Windows terminal does not support coloring, however, macOS and Linux terminals will be colored.

vlsi:
Here's a sample failure: https://github.com/pgjdbc/pgjdbc/actions/runs/2299608243 (try opening test jobs)

vlsi:
I just pushed a commit that would trigger test failure, so we could see how it would look like in the Actions UI.

vlsi:
I think the output is much better, however, it might make sense to exclude successful tests faster than 0.5 from the output.

WDYT?

jlink:
> WDYT?

I personally prefer to only see failures (and maybe warnings) in the output. Everything else is noise to me.

vlsi:
> I personally prefer to only see failures (and maybe warnings) in the output

I agree, however, it is suspicious if a single test takes noticeable time (e.g. 30 seconds).

I've adjusted the threshold of slow tests to 15sec, and the output is cleaner now: https://github.com/jlink/jqwik/runs/6637715794?check_suite_focus=true#step:5:1295

I believe printing tests exceeding 15 sec might help capture cases like accidental algorithmic complexity like in `CombinedIterator`. Of course, `15 sec` is adjustable, and you might have better idea what threshold would work for jqwik better.

buildSrc/src/main/groovy/jqwik.common-configuration.gradle
```diff
@@ -1,3 +1,8 @@
+plugins {
+	// See https://github.com/vlsi/vlsi-release-plugins/blob/master/plugins/gradle-extensions-plugin/README.md
+	id 'com.github.vlsi.gradle-extensions'
+}
+
```
jlink, line 5: Could the syntax from below be used here:
```
project.plugins.apply('com.github.vlsi.gradle-extensions')
```
But I might be confusing different Gradle concepts...

vlsi: Gradle's recommendation is to mention all the plugins in the first `plugins` block, and they recommend `id(..)` syntax (see https://docs.gradle.org/current/userguide/plugins.html#sec:plugins_block ).

At the same time, applying plugins via `apply` call is discouraged (see https://docs.gradle.org/current/userguide/plugins.html#sec:old_plugin_application )

I do not follow why you suggest using deprecated techniques, especially when keeping #281 in mind.

In any case, I do not care regarding the exact syntax you prefer, and it is hard for me to understand the reasons you prefer `project.plugins.apply`, so feel free to adjust the code the way you like.

If you absolutely want me to do the change, of course, I can do it, however, it looks like a waste of time from my point of view.

jlink: I wonder if the trade-offs are different in shared configurations?

Anyway, at least it should be consistent. Let's ask the creator of the shared configs:
@tmohme Why did you use `project.plugins.apply` instead of `plugins { id '...' }` in https://github.com/jlink/jqwik/pull/282?

vlsi: I believe, that was by mistake. If you check https://github.com/jlink/jqwik/pull/282/files, you'll see that the PR added a lot `plugins {..}` calls, and `project.plugins.apply` was just in two places. They beg for moving into `plugins{...}`

jlink: In that case, I'd vote for moving everything to `plugins{..}` in this PR.

tmohme: To be honest, I don't remember for sure why I used the `project.plugins.apply` syntax in the convention plugins (as opposed to the plugins-block syntax everywhere else).
Most likely I've seen it somewhere in the gradle docs, maybe there was a limitation some time ago, maybe it was a simple mistake  . . .
Anyway, I don't see any advantages in keeping the `project.plugins.apply` syntax and also would also recommend the `plugins {...}` syntax now.

jlink: @tmohme Thanks for clarifying!

jlink:
@vlsi You'll have to rebase since I switched to plugins DSL in shared configurations

PR #352, opened 2022-05-28
ci: use GitHub workflow concurrency feature to cancel stale PR jobs

vlsi:
It will automatically cancel the past CI jobs in case PR is updated with a new commit

---

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -9,6 +9,13 @@ on:
     branches:
       - '*'
 
+concurrency:
+  # On main, we don't want any jobs cancelled so the sha is used to name the group
+  # On PR branches, we cancel the job if new commits are pushed
+  # More info: https://stackoverflow.com/a/68422069/253468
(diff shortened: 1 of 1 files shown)


vlsi:
In theory, it requires at least one build in `main` beach (`burrunan/gradle-cache-action` is smart enough to keep reuse gradle and dependency cache between `main` branch and `PR` builds), however, even now the results are impressive:

The current `jqwik` ci: https://github.com/jlink/jqwik/actions/runs/2401112410
baseline ci duration

The suggested `burrunan/gradle-cache-action`: https://github.com/jlink/jqwik/actions/runs/2401558661
burrunan/gradle-cache-action is much faster

The thing is that `burrunan/gradle-cache-action` implements Gradle Build Cache via GitHub Actions Cache API.
The cache from PR does not pollute the cache in `main` branch, however, the items cached during `main` build could be reused in the `PR` builds

jlink:
@Thanks

PR #358, opened 2022-06-03
Remove FeatureExtractor#applySafe to let the error propagate

vlsi:
## Overview

This PR removes `FeatureExtractor#applySafe`.

### Details

The current implementation might hide exceptions in the user-provided code, so it might make error understanding harder than it could be.

---

### DIFF
--- a/engine/src/test/java/net/jqwik/api/ListArbitraryTests.java
+++ b/engine/src/test/java/net/jqwik/api/ListArbitraryTests.java
@@ -97,7 +97,7 @@ class ListArbitraryTests {
 		ListArbitrary<Integer> listArbitrary =
 			Arbitraries.integers().between(1, 1000).injectNull(0.5)
 					   .list().ofMaxSize(20)
-					   .uniqueElements(i -> i % 100);
+					   .uniqueElements(i -> i == null ? null : i % 100);
 
 		RandomGenerator<List<Integer>> generator = listArbitrary.generator(1000, true);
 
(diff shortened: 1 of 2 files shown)


engine/src/test/java/net/jqwik/api/ListArbitraryTests.java
```diff
@@ -97,7 +97,7 @@ void uniqueElementsWithNull(@ForAll Random random) {
 		ListArbitrary<Integer> listArbitrary =
 			Arbitraries.integers().between(1, 1000).injectNull(0.5)
 					   .list().ofMaxSize(20)
-					   .uniqueElements(i -> i % 100);
+					   .uniqueElements(i -> i == null ? null : i % 100);
```
vlsi, line 100: `i` is `Integer i`.
Since the generator has `injectNull`, I believe it is reasonable that the user must handle `null` case, and they must not attempt to unbox `null` value.

jlink: `i` is `Integer` because `int` cannot be used as a type parameter.

I'd like to keep the above behaviour unchanged.

vlsi: Well. The issue with jqwik is that it hides an exception from a userprovided code.
In other words, consider `.uniqueElements(i -> computeNameOf(i));`
If `computeNameOf` throws NPE, then jqwik would hide that, and the users won't detect the error.

I believe it is absolutely unexpected and unexpected behavior.

jlink: Let me think about that for a while.

PR #368, opened 2022-06-27
Keep the tail of the chain in ShrinkableChainShrinker#replaceRangeByShrunkRange

vlsi:
Previously replaceRangeByShrunkRange kept only the beginning of the chain so the shrinking candidate was not really workable.

I added `for (int i = endIndex + 1; i < iterations.size(); i++)` to append the tail of the chain.


vlsi:
It looks like the approach was like "discard all the actions after any changed one"

I think that won't really work for non-trivial cases, however, changing that would require more changes

PR #414, opened 2022-11-23
Fix typo in provides net.jqwik.api.state.ActionChain.ActionChainFacade

mkleczek:
Not sure how it is possible as it should not compile but it seems there is a typo in provider class name: net.jqwik.engine.facades.ActionChainFacadeImple (note the last letter 'e'). It causes error when using jqwik stateful testing in modular environment:
java.util.ServiceConfigurationError: net.jqwik.api.state.ActionChain$ActionChainFacade: Provider net.jqwik.engine.facades.ActionChainFacadeImple not found.

## Overview

_Formulate the problem / feature your pull request is aiming at_

### Details

_Give any necessary details about your code that is not obvious from the code itself._

_List open issues and remaining problems with your solution._

---


jlink:
@mkleczek Thanks! Well, anything is possible ;-)

Issue #363, opened 2022-06-19
Problem space size vs seed size


vlsi:
## Testing Problem

Problem space might exceed 64bits quite fast, so some of the cases are inherently impossible to represent and catch with jqwik.

For instance, if I want to verify `a + b == b + a` where `a` and `b` are `long` variables, then the problem space is `128bits`, and `64bit` seed is not enough. It will miss "a lot" of cases, and it will never be able to generate them no matter how much time we allocate for the testing.
Of course, "edge case" injection solves that to a degree.

## Discussion

Do you know if using a bigger random state is known to affect property-test quality?

/cc @inponomarev

jlink:
As far as I know, more than 64 bits means that RNG algorithms get measurably slower since some ops can no longer be done with a single machine op. If that will result in a measurably slower generation of values probably depends on the number of required generation steps. In the standard cases I'd doubt that it has an effect.

As for test quality, I'd assume that it won't effect it since value distribution should more or less be the same for reasonable pseudo-random generation algorithms.

vlsi:
See https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/random/package-summary.html#algorithms, https://docs.oracle.com/en/java/javase/17/core/pseudorandom-number-generators.html#GUID-6946463F-C64B-4361-AB53-0D52FB279DE0

> Large Permutations
> For applications that generate large permutations, consider a generator whose period is much larger than the total number of possible permutations; otherwise, it will be impossible to generate some of the intended permutations. For example, if the goal is to shuffle a deck of 52 cards, the number of possible permutations is 52! (52 factorial), which is approximately 2^225.58, so it may be best to use a generator whose period is roughly 2^256 or larger, such as L64X256MixRandom, L64X1024MixRandom, L128X256MixRandom, or L128X1024MixRandom.

The current `net.jqwik.engine.SourceOfRandomness#newRandom(long)` is able to generate only `2.287... × 10^-49 == 2^64/52!` of all card shuffles.

Judging by https://issues.apache.org/jira/browse/RNG-168?focusedCommentId=17520192&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17520192, the difference is not really dramatic.

I'm inclined that something like L64X1024MixRandom or L128X1024MixRandom might be valuable for jqwik: it should not be that hard to maintain, it should have pretty much the same performance, and it would allow users to test more permutations than the current random generator.

[...]

PR #424, opened 2022-12-10
WIP: use L128X1024Mix for random generator

vlsi:
## Overview

See https://github.com/jlink/jqwik/issues/363

This is more like work-in-progress, however, it looks like the changes are not really big.

Basically:
* `long seed` -> `JqwikRandomState seed`
*  `Random` -> `JqwikRandom`
* The seed looks like in base36. It takes 198 chars, it is easy to select by double-clicking. If we add upper-case letters, then it would be 172 chars, which is not a significant improvement, however, we won't be able to reuse Java's standard base36 encode-decode. `seed = 1piidb7t9rfnx31xty9qfxo1jlx1qi9wkegpnxl5w4geilda60j7rbjntetpvuuveabbeejzi02hl4pmwmjaoe61mv8iny98hlgfmgq36fqiw9kdpeocp8qhvj84j56l7g1xhlxlfkr7sne536w2d0rvmksz1ze3gd0y3y3jwfa7q3m3myaj6qi3wwmxdi5mdig2jmy1a3jwb58g4i95aow5h6hodvkom8i9xeuafi4ug009s2maxqyele0gzcz4ykihtvprq7d059n29lpy3hb5qiz29| random seed to reproduce generated values`

There's a possibility to use `net.jqwik.api.Random`, however, I am afraid it would clash with `java.util.Random`.

I'm not sure if it is worth making a distinction between the seed (the one you use to create a random generator), and the state of the generator.

Commons-rng allows to export the state of a random generator to an object, however, it does not provide cross-version compatibility for that state. In other words we can't use saveState/restoreState for `@Property(seed="...")`

At the same time, if we use .split(), .jump() for splitting the generator instead of nextSeed=random.nextLong(), then the only thing we can have is the internal, non-serializable state of the newly split generator.
We can't really infer "the values we should pass as byte[]seed in order to produce the same generator, as we got after the split"

I would try adding both seed (optional) and state (optional, computed from seed if missing) to an opaque interface JqwikRandomState. Then we could both have cross-version seed values and use random.split() in the test engine.

WDYT?

---

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -37,4 +37,4 @@ Use _jqwik_ itself for all tests and properties.
 
 Use _AssertJ_ for non trivial assertions.
 
-Use `@ForAll Random random` parameter if you need a random value. 
+Use `@ForAll JqwikRandom random` parameter if you need a random value. 
--- a/api/build.gradle
+++ b/api/build.gradle
@@ -53,4 +53,6 @@ signing {
 dependencies {
 	api("org.opentest4j:opentest4j:${opentest4jVersion}")
 	api("org.junit.platform:junit-platform-commons:${junitPlatformVersion}")
+	api(platform("org.apache.commons:commons-rng-bom:1.5"))
+	api("org.apache.commons:commons-rng-client-api")
 }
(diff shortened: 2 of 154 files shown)


vlsi:

Issue #423, opened 2022-12-10
Move to a jumpable random generator to avoid statistical depenency between map, flatMap, chain, etc calls


vlsi:
## Testing Problem

Currently, jqwik produces too many duplicate values for the input.

I believe the root cause is `nextSeed = random.nextLong()` pattern that is used in several key places.

For instance, below is a reproducer for `long nextSeed = random.nextLong();` in `net.jqwik.engine.properties.state.ShrinkableChain.ChainIterator#nextTransformer`: 

https://github.com/jlink/jqwik/blob/7a37d8dd5c66616320390e5c284a2e8eac6023fd/engine/src/main/java/net/jqwik/engine/properties/state/ShrinkableChain.java#L170-L172

The problem with `nextSeed = random.nextLong()` is that both random generators (the original and the one produced with `SourceOfRandomness.newRandom(nextSeed)`) would follow the same sequence.

For instance, try adding the following code to `ActionChainArbitraryTests`:

Frankly speaking, I find the syntax verbose, however, let it put aside for a while.

The idea of the test is that it creates `Set<Integer>` for the state, and performs one of the following actions:
1) clear the set
2) add a random integer between 1, Integer.MAX_VALUE there

Since I generate 30 actions, I would expect all the values to be unique

```java
	@Property(tries = 10000)
	void duplicatesInChain(@ForAll Random random) {
		ActionChainArbitrary<Set<Integer>> chains =
			ActionChain.<Set<Integer>>startWith(HashSet::new)
			   .addAction(
				   1, 
				   Action.<Set<Integer>>when(set -> !set.isEmpty())
					   .just(Transformer.mutate("clear", Set::clear))
			   )		
			   .addAction(
				   5, 
				   (Action.Dependent<Set<Integer>>) state ->
						Arbitraries.integers()
						   .between(1, 10000000)
						   .withDistribution(RandomDistribution.uniform())
						   .withoutEdgeCases()
						   .map(i -> {
									if (state.contains(i)) {
										// System.out.println("got duplicate value " + i);
										Statistics.label("value").collect("duplicate");
										return Transformer.noop();
									} else {
										Statistics.label("value").collect("unique");
										return Transformer.mutate("add " + i, set -> set.add(i));
									}
								}
						   )
			   )
			   .withMaxTransformations(30);

		ActionChain<Set<Integer>> chain = TestingSupport.generateFirst(chains, random);
		chain.run();
	}
```

The output is

```
[...]

PR #425, opened 2022-12-11
Add mix64 to XORShiftRandom constructor to make newRandom(random.nextLong()) better

vlsi:
See https://github.com/jlink/jqwik/issues/423

### DIFF
--- a/engine/src/main/java/net/jqwik/engine/SourceOfRandomness.java
+++ b/engine/src/main/java/net/jqwik/engine/SourceOfRandomness.java
@@ -62,10 +62,20 @@ public class SourceOfRandomness {
 		}
 
 		private XORShiftRandom(long seed) {
-			if (seed == 0l) {
-				throw new IllegalArgumentException("0L is not an allowed seed value");
+			this.seed = mix64(seed);
+			if (this.seed == 0) {
+				// 0 is invalid for XorShift seed, so we set it to a non-zero value
+				this.seed = 0xbf58476d1ce4e5b9L;
 			}
-			this.seed = seed;
+		}
+
+		/**
+		* See <a href="http://zimbry.blogspot.com/2011/09/better-bit-mixing-improving-on.html">Better Bit Mixing - Improving on MurmurHash3's 64-bit Finalizer</a>
+		*/
+		private static long mix64(long z) {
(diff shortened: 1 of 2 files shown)


engine/src/main/java/net/jqwik/engine/SourceOfRandomness.java
```diff
@@ -65,7 +65,16 @@ private XORShiftRandom(long seed) {
...
+			this.seed = mix64(seed);
+		}
+
+		/**
+		* See <a href="http://zimbry.blogspot.com/2011/09/better-bit-mixing-improving-on.html">Better Bit Mixing - Improving on MurmurHash3's 64-bit Finalizer</a>
+		*/
+		private static long mix64(long z) {
+			z = (z ^ (z >>> 30)) * 0xbf58476d1ce4e5b9L;
+			z = (z ^ (z >>> 27)) * 0x94d049bb133111ebL;
+			return z ^ (z >>> 31);
```
jlink, line 77: Any reason for choosing this exact mode of scrambling and not just adding 1? Just being curious...

vlsi: Frankly speaking, Copilot suggested me that function for scrambling the value.

I put `this.seed = `, and it autocomplete mix64 literally, except the reference link.
Then I searched the constants online, and I found the corresponding article, so I believe the constants are both good, and ok license-wise.

vlsi: One more thing: it might be that the value will be zero after `mix64`, so it looks like we need to do something about "0L is not an allowed seed value".

What if we replace the result of `mix64` with `0xbf58476d1ce4e5b9L` if it turns out to be `0`?
Then it would have a non-zero starting value, and it would avoid unexpected "all zeroes caused by 0 after mix64"

jlink: Is the probability higher than 1/2^64? Not sure if it's worth the performance hit and the extra complexity.

vlsi: I believe the performance hit is virtually zero, and the complexity is not there as well: `0` check was there anyway, so I replace the exception with a reasonable value to start with.

Issue #429, opened 2022-12-14
Cache LifecycleMethods results: findBeforeTryMethods, findAfterTryMethods, etc


vlsi:
## Testing Problem

`LifecycleMethods` consumes significant CPU time performing the same reflective lookup.

## Suggested Solution

Cache the results via https://docs.oracle.com/javase/7/docs/api/java/lang/ClassValue.html or something like that.

## Profiling results

Allocation:

<img width="1405" alt="allocation profiling of LifecycleMethods" src="https://user-images.githubusercontent.com/213894/207538621-1897ff4c-332c-4b61-bfbd-6644a5415677.png">

CPU:

<img width="1391" alt="CPU profiling of LifecycleMethods" src="https://user-images.githubusercontent.com/213894/207538640-e0102262-d508-4dde-8f79-f8e80a943e05.png">

vlsi:
Sample fix:

```diff
+       private static final ClassValue<List<Method>> findBeforeTryMethods = new ClassValue<List<Method>>() {
+               @Override
+               protected List<Method> computeValue(Class<?> testClass) {
+                       return findMethods(testClass, false, true, BeforeTry.class, HierarchyTraversalMode.TOP_DOWN);
+               }
+       };
+
        static List<Method> findBeforeTryMethods(Class<?> testClass) {
-               return findMethods(testClass, false, true, BeforeTry.class, HierarchyTraversalMode.TOP_DOWN);
+               return findBeforeTryMethods.get(testClass);
        }

+       private static final ClassValue<List<Method>> findAfterTryMethods = new ClassValue<List<Method>>() {
+               @Override
+               protected List<Method> computeValue(Class<?> testClass) {
+                       return findMethods(testClass, false, true, AfterTry.class, HierarchyTraversalMode.BOTTOM_UP);
+               }
+       };
+
        static List<Method> findAfterTryMethods(Class<?> testClass) {
-               return findMethods(testClass, false, true, AfterTry.class, HierarchyTraversalMode.BOTTOM_UP);
+               return findAfterTryMethods.get(testClass);
        }
```

WDYT?

It improves shrink time from ~60s to 30sec for case like https://github.com/jlink/jqwik/issues/428#issue-1492125145 (increasing `.withMaxTransformations(12)` makes the case WAY longer to execute)

PR #430, opened 2022-12-14
Cache reflective lookup in LifecycleMethods for before/after try, property, and container methods

vlsi:
Fixes https://github.com/jlink/jqwik/issues/429

---


jlink:
@vlsi Thanks for that one. I learned a new thing with `ClassValue`.

PR #432, opened 2022-12-14
Optimize .flatMap(... -> Stream.of(..)) with .map(...) in ShrinkableChainShrinker

vlsi:
This is a small optimization that mainly makes code easier to follow, however, it does improve perf from 28 to 26 sec

---


vlsi:
I believe CI failure has nothing to do with the change.

PR #434, opened 2022-12-14
Optimize BigIntegerGrower#toLong: use instanceof instead of catching ClassCastException

vlsi:
Fixes https://github.com/jlink/jqwik/issues/433

---

I hereby agree to the terms of the [jqwik Contributor Agreement](https://github.com/jlink/jqwik/blob/master/CONTRIBUTING.md#jqwik-contributor-agreement).

PR #435, opened 2022-12-14
Optimize ShrinkingDistance: reduce long[] allocations when computing distance for collections

vlsi:
## Overview

`ShrinkingDistance#forCollection` and `combine` are used a lot, and they might easily account for lots of time spent in re-allocating long arrays.

Sample:

ShrinkingDistance in allocation hotspot

---

### DIFF
--- a/api/src/main/java/net/jqwik/api/ShrinkingDistance.java
+++ b/api/src/main/java/net/jqwik/api/ShrinkingDistance.java
@@ -4,6 +4,7 @@ import java.util.*;
 import java.util.stream.*;
 
 import org.apiguardian.api.*;
+import org.jetbrains.annotations.*;
 
 import static org.apiguardian.api.API.Status.*;
 
@@ -12,6 +13,8 @@ public class ShrinkingDistance implements Comparable<ShrinkingDistance> {
 
 	@API(status = MAINTAINED, since = "1.2.0")
 	public static final ShrinkingDistance MAX = ShrinkingDistance.of(Long.MAX_VALUE);
+
+	private static final ShrinkingDistance MIN = ShrinkingDistance.of(0);
 
 	private final long[] distances;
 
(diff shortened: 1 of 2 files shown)


api/src/main/java/net/jqwik/api/ShrinkingDistance.java
```diff
@@ -22,20 +25,44 @@ public static ShrinkingDistance of(long... distances) {
...
+		for (Shrinkable<T> element : elements) {
+			long[] next = element.distance().distances;
+			if (data == null) {
+				data = Arrays.copyOf(next, next.length);
+				continue;
+			}
+			data = sumUpArrays(data, data, next);
+
+		}
+		return new ShrinkingDistance(concatArrays(new long[]{elements.size()}, data));
```
vlsi, line 42: This could probably be removed by adding `startOffset` argument to `sumUpArrays`.

jlink:
Without having gone through all the details of the expanded code, I'm wondering how much performance win there is.
Do you have a test case available I can use to get a feeling for it?

vlsi:
I think I profiled this in jqwik test suite. It could be thr same test as https://github.com/jlink/jqwik/pull/434#issuecomment-1350980918

In my test I have collections with up to 1000 elements, so computing distance for those elements would cause 1000 re-allocations of double[] distance arrays which would definitely be a waste of time and resources.

An alternative approach would be creating DistanceBuilder that would contain mutable double[] (and pass it to the resulting Distance object). It could probably make code easier to follow, however, I heard you do not like builders ( ;) ), and the code is on the hot path.

jlink:
@vlsi I'll probably not use the PR verbatim but use it as inspiration if that's ok with you.

vlsi:
Go for it if it resolves the CPU and heap allocation issues.

jlink:
@vlsi Current code in main (deployed to `1.7.2-SNAPSHOT`) could prevent one or two temporary list or array creations, but feels like a reasonable compromise between readable code and performance. Please test when you have time. Many thanks in advance.

vlsi:
It is both hard to review and comment since you made several incoherent changes.


```java
	public static long[] sumUp(List<long[]> listOfArrays) {
		int maxDistanceSize = listOfArrays.stream().mapToInt(s -> s.length).max().orElse(0);
		long[] summedUpArray = new long[maxDistanceSize];
		Arrays.fill(summedUpArray, 0);
		for (long[] array : listOfArrays) {
			for (int i = 0; i < summedUpArray.length; i++) {
				summedUpArray[i] = plusWithoutOverflowAt(summedUpArray, array, i);
			}
		}
		return summedUpArray;
	}
```

`listOfArrays.stream().mapToInt(s -> s.length).max().orElse(0);` allocates memory even though it is on a critical path.
A plain for-loop would be just fine there.

`Arrays.fill(summedUpArray, 0);` Java language specification guarantees that all allocated arrays have default values. So `Arrays.fill` after `new long` is pure noise.

```java
		for (long[] array : listOfArrays) {
			for (int i = 0; i < summedUpArray.length; i++) {
```

It would be more efficient to swap the order of loops (cache locality would be way better when scanning through arrays just once).

`plusWithoutOverflowAt`: does not handle overflow properly in case one of the input values is negative. The method is in public API, and such naming would induce surprising behaviors.

`LongArraysSupport#at(long[] array, int i)` is in public API, and it surprisingly replaces non-existing elements with `0`. The method should be private, or it should be renamed.

----

LongArraysSupport#concatenate: `listOfArrays.stream().mapToInt(s -> s.length).sum();`. It would be better to just use for loop there.
Well, for `concatenate`, there's a stream-based implementation, however, I have not benchmarked how it performs.
In any case, it looks surprising to see both `stream` and `for` in the same critical code path.

```java
listOfArrays.stream().flatMapToLong(LongStream::of).toArray()
```

jlink:
@vlsi Thanks for all the feedback. I'll wrap it up for the time being.

Issue #428, opened 2022-12-12
ActionChain shrink should be called only with state objects that correspond to the relevant state position


vlsi:
## Testing Problem

It looks like `.shrink()` can be called on a wrong state instance.

The key idea behind the test:
1. There are actions to `add` int to a set, and `clear` the set
1. The very last action attempts to print a value out of a `set`. Note: it selects the value to print based on the `state.set`, so I assume it should select valid values only.

If the value is not present in the "print transformer", then I believe it is illegal.

The following test throws `IllegalStateException`

```java
static class SetMutatingChainState {
	final List<String> actualOps = new ArrayList<>();
	final Set<Integer> set = new HashSet<>();
	boolean hasPrints = false;

	@Override
	public String toString() {
		return "set=" + set + ", actualOps=" + actualOps;
	}
}

@Property(shrinking = ShrinkingMode.FULL, seed = "7077187739734332001")
void chainActionsAreProperlyDescribedEvenAfterChainExecution(@ForAll("setMutatingChain") ActionChain<SetMutatingChainState> chain) {
	chain = chain.withInvariant(
		state -> {
			if (state.hasPrints) {
				assertThat(state.actualOps).hasSizeLessThan(5);
			}
		}
	);

	SetMutatingChainState finalState = chain.run();

	assertThat(chain.transformations())
		.describedAs("chain.transformations() should be the same as the list of operations in finalState.actualOps, final state is %s", finalState.set)
		.isEqualTo(finalState.actualOps);
}

@Provide
public ActionChainArbitrary<SetMutatingChainState> setMutatingChain() {
	return
		ActionChain
			.startWith(SetMutatingChainState::new)
			.withAction(
				1,
				Action.just("clear anyway", state -> {
					state.actualOps.add("clear anyway");
					state.set.clear();
					return state;
				})
			)
			// Below actions depend on the state to derive the transformations
			.withAction(
				1,
				(Action.Dependent<SetMutatingChainState>)
					state ->
						Arbitraries
							.just(
								state.set.isEmpty()
									? Transformer.noop()
									: Transformer.<SetMutatingChainState>mutate("clear " + state.set, set -> {
									state.actualOps.add("clear " + set.set);
									state.set.clear();
								})
							)
			)
			.withAction(
				4,
				(Action.Dependent<SetMutatingChainState>)
					state ->
						Arbitraries
							.integers()
							.between(1, 10)
							.map(i -> {
									 if (state.set.contains(i)) {
[...]

Issue #338, opened 2022-05-19
Shrinking stateful chain yields OutOfMemoryError


vlsi:
## Testing Problem

Shrinking stateful chain causes the following.
I'll attach details from the heap dump soon

```
java.lang.OutOfMemoryError: Java heap space
Dumping heap to java_pid6286.hprof ...
Heap dump file created [920534673 bytes in 4.318 secs]
MappingApply2Properties > mapping change works STANDARD_ERROR
    мая 19, 2022 8:05:21 PM net.jqwik.engine.JqwikTestEngine execute
    SEVERE: Java heap space
    java.lang.OutOfMemoryError: Java heap space
    	at java.base/java.util.ArrayList.iterator(ArrayList.java:977)
    	at net.jqwik.engine.support.combinatorics.CombinedIterator.<init>(CombinedIterator.java:21)
    	at net.jqwik.engine.support.combinatorics.CombinedIterator.restIterator(CombinedIterator.java:28)
    	at net.jqwik.engine.support.combinatorics.CombinedIterator.<init>(CombinedIterator.java:22)
    	at net.jqwik.engine.support.combinatorics.CombinedIterator.restIterator(CombinedIterator.java:28)
    	at net.jqwik.engine.support.combinatorics.CombinedIterator.<init>(CombinedIterator.java:22)
    	at net.jqwik.engine.support.combinatorics.CombinedIterator.restIterator(CombinedIterator.java:28)
    	at net.jqwik.engine.support.combinatorics.CombinedIterator.<init>(CombinedIterator.java:22)
    	at net.jqwik.engine.support.combinatorics.CombinedIterator.restIterator(CombinedIterator.java:28)
    	at net.jqwik.engine.support.combinatorics.CombinedIterator.<init>(CombinedIterator.java:22)
    	at net.jqwik.engine.support.combinatorics.CombinedIterator.restIterator(CombinedIterator.java:28)
    	at net.jqwik.engine.support.combinatorics.CombinedIterator.<init>(CombinedIterator.java:22)
    	at net.jqwik.engine.support.combinatorics.CombinedIterator.restIterator(CombinedIterator.java:28)
    	at net.jqwik.engine.support.combinatorics.CombinedIterator.<init>(CombinedIterator.java:22)
    	at net.jqwik.engine.support.combinatorics.CombinedIterator.restIterator(CombinedIterator.java:28)
[...]

PR #436, opened 2022-12-14
Cache several shrink results for every chain step, so it does not re-access state model state during shrink phase

vlsi:
## Overview

See "a" in https://github.com/jlink/jqwik/issues/428#issuecomment-1350497917

a) Sample (and keep) several .shrink() results for each action before the action is executed.
Pro: we would have valid shrink results (the shrink result would strictly correspond to the state)
Cons: it would spend time and memory on keeping .shrink() results even in case the chain succeeds.
Cons: it does not fully fit "try many different shrink types, order them, and start with the most promising ones" since we would be able to keep only a few shrink results for each action.

### Details

The number of cached shrinkables could be configured via `jqwik.eagerChainShrinkSamples` system property. I default it to 2.

---

### DIFF
--- a/engine/src/main/java/net/jqwik/engine/properties/state/ShrinkableChainIteration.java
+++ b/engine/src/main/java/net/jqwik/engine/properties/state/ShrinkableChainIteration.java
@@ -2,16 +2,63 @@ package net.jqwik.engine.properties.state;
 
 import java.util.*;
 import java.util.function.*;
+import java.util.stream.*;
 
 import net.jqwik.api.*;
 import net.jqwik.api.state.*;
 
-import org.jetbrains.annotations.*;
-
 class ShrinkableChainIteration<T> {
+	// Larger values might improve shrink quality, however, they increase the shrink space, so it might increase shrink duration
+	private final static int NUM_SAMPLES_IN_EAGER_CHAIN_SHRINK = Integer.getInteger("jqwik.eagerChainShrinkSamples", 2);
+
+	static class ShrinkableWithEagerValue<T> implements Shrinkable<T> {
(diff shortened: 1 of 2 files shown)


engine/src/main/java/net/jqwik/engine/properties/state/ShrinkableChainIteration.java
```diff
@@ -2,11 +2,61 @@
...
+
+	static class EagerShrinkable<T> extends ShrinkableWithEagerValue<T> {
+		private final List<Shrinkable<T>> shrinkResults;
+
+		EagerShrinkable(Shrinkable<T> base, int numSamples) {
+			super(base);
+			this.shrinkResults =
+				base.shrink()
+					.sorted(Comparator.comparing(Shrinkable::distance))
+					.limit(numSamples)
```
vlsi, line 49: This triggers OOM in case `.shrink()` stream is big. See https://github.com/jlink/jqwik/issues/338#issuecomment-1351637147

vlsi:
I tried re-enabling shrink for my action chain, and it turns out to spend too much time shrinking (too many variations? too many restrictions in the current shrink approach like "after shrink, all the subsequent state-accessing actions must be discarded"?)

I tried this change with my current tests. and the property fails after ~8 minutes with shrink=OFF producing ~15 actions. However, if I enable shrinking and re-run with the same seed, then it can't converge after 40-50 minutes, so I am inclined to just go for shrink=OFF and get the failure faster.

I consider implementing toString-like methods that would print **test code** instead of "list of actions". Then I would be able to "shrink" the test manually by commenting-out irrelevant bits of the generated test code.

vlsi:

PR #491, opened 2023-06-17
Improved performance of @UniqueElements(by=NOT_SET)`

DirkToewe:
## Overview

This PR aims at significantly improving the performance of `@UniqueElements(by=NOT_SET)`, mainly by replacing calls to `uniqueElements(x->x)` by `uniqueElements()` wherever possible.

### Details

If we look at the following two property tests:

```java
@PropertyDefaults( tries = 100_000 )
public class UniqueElementsTest
{
  @Property void testA( @ForAll @UniqueElements int[] a ) {
    Arrays.sort(a);
    for( int i=1; i < a.length; i++ )
      assert a[i-1] <= a[i];
  }

  @Provide Arbitrary<int[]> arrays() {
    return Arbitraries.integers().array(int[].class).uniqueElements();
  }

  @Property void testB( @ForAll("arrays") int[] a ) {
    Arrays.sort(a);
    for( int i=1; i < a.length; i++ )
      assert a[i-1] <= a[i];
  }
}
```

On my machine, it takes 20 seconds for `testA` to finish while `testB` only takes 2 seconds. The problem becomes much more emphasized with larger array sizes.

---

### DIFF
--- a/engine/src/main/java/net/jqwik/engine/properties/configurators/UniqueElementsConfigurator.java
+++ b/engine/src/main/java/net/jqwik/engine/properties/configurators/UniqueElementsConfigurator.java
@@ -1,6 +1,7 @@
 package net.jqwik.engine.properties.configurators;
 
 import java.util.*;
+import java.util.concurrent.*;
 import java.util.function.*;
 import java.util.stream.*;
 
@@ -9,7 +10,6 @@ import net.jqwik.api.arbitraries.*;
 import net.jqwik.api.configurators.*;
 import net.jqwik.api.constraints.*;
 import net.jqwik.api.providers.*;
-import net.jqwik.api.support.*;
 import net.jqwik.engine.support.*;
 
 @SuppressWarnings("unchecked")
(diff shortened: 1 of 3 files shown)


engine/src/main/java/net/jqwik/engine/properties/configurators/UniqueElementsConfigurator.java
```diff
@@ -63,13 +63,28 @@ private <T> List<T> toList(Iterator<T> i) {
...
-		Set<Object> set = list.stream().map(extractor).collect(CollectorsSupport.toLinkedHashSet());
-		return set.size() == list.size();
+	private static <C extends Collection<? extends Object>> Predicate<C> isUnique(UniqueElements uniqueness) {
+		Class<? extends Function<?, Object>> extractorClass = uniqueness.by();
+		if (extractorClass.equals(UniqueElements.NOT_SET.class)) {
+			// intentionally use `items.getClass().equals(HashSet.class)`
+			// instead of `items instanceof HashSet`, because subclasses
+			// of HashSet may break Set conventions.
+			return items -> items.getClass().equals(HashSet.class)
+								|| items.size() == items.stream().distinct().count();
```
vlsi, line 73: This is a suboptimal way to check uniqueness. You can stop as soon as you detect the first non-unique element.

At the same time, elements in hashset are always unique, aren't they?

DirkToewe: Oh You're right, it would be much faster to stop early. I will fix that.

Yes that's why added the `HashSet` clause. It might speed things up sometimes.

DirkToewe: Changed the logic. Now it should stop early.

jlink:
@DirkToewe As soon as you deem it ready, just rebase it, so that I can merge it.

engine/src/main/java/net/jqwik/engine/properties/configurators/UniqueElementsConfigurator.java
vlsi: Nice. There's one more case doing `stream.distinct.count`: https://github.com/jqwik-team/jqwik/blob/124d7af3d56fabfb5b5bf182c339cb3bff3825f9/engine/src/main/java/net/jqwik/engine/properties/FeatureExtractor.java#L32
I wonder if it makes sense to unify them.

WDYT?

DirkToewe:
I did a soft reset and re-committed the changes. All merge conflicts should be gone now.

The PR now also includes the change to  `FeatureExtractor::areUnique` suggested by @vlsi. It felt like it fits the theme of the PR. I hope that's okay.

jlink:
Published in 1.7.4-SNAPSHOT

PR #500, opened 2023-07-10
Issue 449

jlink:
## Overview

Fixes issue #449 and other variance problems of `TypeUsage.canBeAssignedTo(..)`.

---

### DIFF
--- a/engine/src/main/java/net/jqwik/engine/support/types/TypeUsageImpl.java
+++ b/engine/src/main/java/net/jqwik/engine/support/types/TypeUsageImpl.java
@@ -397,11 +397,7 @@ public class TypeUsageImpl implements TypeUsage, Cloneable {
 
 	@Override
 	public boolean canBeAssignedTo(TypeUsage targetType) {
-		// TODO: Recursive calls to this method will lead to a stack overflow in recursive types
-		// e.g. T extends Comparable<T>
-		// This happens to the new anySupertypeCanBeAssignedTo(targetType) call
-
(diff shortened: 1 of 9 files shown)


engine/src/main/java/net/jqwik/engine/support/types/TypeUsageImpl.java
```diff
@@ -661,6 +689,14 @@ public List<TypeUsage> getInterfaces() {
 		return toTypeUsages(getRawType().getGenericInterfaces());
 	}
 
+	@Override
+	public List<TypeUsage> getSuperTypes() {
+		List<TypeUsage> supertypes = new ArrayList<>();
+		getSuperclass().ifPresent(supertypes::add);
```
vlsi, line 695: This seems to retrieve at most one super type rather than all of them

jlink: There can only be one superclass. What am I missing?
It's only about the _direct_ supertypes not indirect ones. Those are being picked when diving deeper.

vlsi: In that case `getDirectSuperTypes` would make the code easier to follow

jlink: Since it's also `getSuperclass/getInterfaces` and not `getDirectSuperclass` that would be inconsistent IMO.

Issue #499, opened 2023-07-07
Domains inject parameterized values incorrectly


SimY4:
## Testing Problem

The setup like this:
```java
@Domain(MyDomain.class)
@Domain(DomainContext.Global.class)
class TestTest {
  @Property
  boolean test(@ForAll Predicate<Number> p1, @ForAll Predicate<Number> p2, @ForAll Number n) {
    return p1.or(p2).test(n) == (p1.test(n) || p2.test(n));
  }
}

class MyThing implements Predicate<CharSequence> {
  public boolean test(CharSequence cs) { return true; }
}

class MyDomain extends DomainContextBase {
  @Provide
  Arbitrary<MyThing> myThingArbitrary() {
    return Arbitraries.just(new MyThing());
  }
}
```

will produce:

```
  java.lang.ClassCastException:
    class java.lang.Long cannot be cast to class java.lang.CharSequence (java.lang.Long and java.lang.CharSequence are in module java.base of loader 'bootstrap')
```

but if you remove domain and provide `MyThing` right in test class - no problem.

jlink:
@SimY4 Interesting catch. Will look into it.

jlink:
OK. It's a bug, but sadly one I've tried to fix a couple of times already without seeing a straightforward solution.

See https://github.com/jqwik-team/jqwik/blob/5055eb3f432e30b416f22a05cebaf7c56a0e3a33/engine/src/main/java/net/jqwik/engine/support/types/TypeUsageImpl.java#L410C1-L417C1

Maybe I'll find the time to reinvestigate. The problem is that type matching is sometimes too loose. And it's very hard to fix this without making other cases fail.

In your case. The following change would fix it, but you probably know that already:
```java
class MyDomain extends DomainContextBase {
	@Provide
	Arbitrary<Predicate<CharSequence>> myThingArbitrary() {
		return Arbitraries.just(new MyThing());
	}
}
```

SimY4:
@jlink yeah, I figured. Not sure why it works for provided in test though. Is it just by chance? I get that it's probably related to MyThing not having any explicit type arguments. Should it find and compare  only matching interfaces between requirements and provided definitions? I.e. I need Predicate<Number> I have MyThing which has Predicate<CharSeq> but Number is not assignable from CharSeq so no.

jlink:
> @jlink yeah, I figured. Not sure why it works for provided in test though. Is it just by chance? 

[...]

PR #505, opened 2023-07-16
Issue499

jlink:
## Overview

Supposed to resolve issue #499 

---

### DIFF
--- a/engine/src/main/java/net/jqwik/engine/support/types/TypeUsageImpl.java
+++ b/engine/src/main/java/net/jqwik/engine/support/types/TypeUsageImpl.java
@@ -404,6 +404,10 @@ public class TypeUsageImpl implements TypeUsage, Cloneable {
 			return canBeAssignedToUpperBounds(this, targetType) &&
 					   canBeAssignedToLowerBounds(this, targetType);
 		}
+		if (this.isTypeVariableOrWildcard()) {
+			return canBeAssignedFromUpperBounds(this, targetType) &&
+					   canBeAssignedFromLowerBounds(this, targetType);
+		}
 		if (primitiveTypeToObject(this.getRawType(), targetType.getRawType()))
 			return true;
 		if (boxedTypeMatches(targetType.getRawType(), this.rawType))
@@ -420,7 +424,7 @@ public class TypeUsageImpl implements TypeUsage, Cloneable {
 				// Sadly this will only work for direct recursion.
 				// Indirect recursion will still lead to a stack overflow.
 				boolean isRecursive = targetType.getTypeArgument(0).getUpperBounds().stream()
-										   .anyMatch(upperBound -> upperBound.isOfType(targetType.getRawType()));
+												.anyMatch(upperBound -> upperBound.isOfType(targetType.getRawType()));
 				if (isRecursive) {
 					return true;
 				}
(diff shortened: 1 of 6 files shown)


engine/src/main/java/net/jqwik/engine/support/types/TypeUsageImpl.java
```diff
@@ -681,18 +641,83 @@ public Optional<TypeUsage> getSuperclass() {
...
 		}
-		return Optional.of(TypeUsage.forType(rawType.getGenericSuperclass()));
+		// Double checking for thread safety
+		if (superclass == null) {
+			synchronized (this) {
+				if (superclass == null) {
+					superclass = createSuperclass();
+				}
+			}
+			superclass = createSuperclass();
```
vlsi, line 651: Double-creating the superclass looks weird

engine/src/main/java/net/jqwik/engine/support/types/TypeUsageImpl.java
```diff
@@ -681,18 +641,83 @@ public Optional<TypeUsage> getSuperclass() {
 		if (rawType.getSuperclass() == null) {
 			return Optional.empty();
 		}
-		return Optional.of(TypeUsage.forType(rawType.getGenericSuperclass()));
+		// Double checking for thread safety
+		if (superclass == null) {
+			synchronized (this) {
```
vlsi, line 646: This is not safe.
See https://shipilev.net/blog/2014/safe-public-construction/#_safe_publication
`superclass` is not `volatile`, so you must not read the value "without a lock" if you need correctness.
Of course, there are exceptions, however, I am pretty sure this is a plain race condition (unsafe publication) since `TypeUsageImpl` is mutable

engine/src/main/java/net/jqwik/engine/support/types/TypeUsageImpl.java
jlink: it is. copy-paster error.

engine/src/main/java/net/jqwik/engine/support/types/TypeUsageImpl.java
```diff
@@ -681,18 +641,83 @@ public Optional<TypeUsage> getSuperclass() {
...
+		}
+		return namedTypeArguments;
 	}
 
 	@Override
 	public List<TypeUsage> getInterfaces() {
-		return toTypeUsages(getRawType().getGenericInterfaces());
+		// Double checking for thread safety
+		if (interfaces == null) {
+			synchronized(this) {
```
vlsi, line 700: This is not safe, see https://shipilev.net/blog/2014/safe-public-construction/#_safe_initialization

engine/src/main/java/net/jqwik/engine/support/types/TypeUsageImpl.java
jlink: volatile is definitely necessary.

engine/src/main/java/net/jqwik/engine/support/types/TypeUsageImpl.java
```diff
@@ -310,6 +310,13 @@ private static List<TypeUsage> extractLowerBoundsForWildcard(WildcardType wildca
 	private HashMap<String, Object> metaInfo = new LinkedHashMap<>();
 	private boolean isNullable = false;
 
+	// Lazy initialization
+	volatile private TypeUsage superclass = null;
+
+	// Lazy initialization
+	volatile private List<TypeUsage> interfaces = null;
```
vlsi, line 317: I would suggest `private volatile`.

See
* https://google.github.io/styleguide/javaguide.html#s4.8.7-modifiers
* https://cr.openjdk.org/~alundblad/styleguide/index-v6.html#toc-modifiers

Issue #329, opened 2022-05-03
Add filtering to combine(...) (combineNotNull?)


vlsi:
## Testing Problem

I combine arbitraries (for state-dependent testing), however, certain values are not acceptable, so I want to filter the outcome of the combination.

For instance, for testing `swap(i, j)` I need two indices. However, I would like to avoid calling `swap` when `i==j`.

```kotlin
combine(ints(), ints()) { i, j ->
    if (i == j) {
        ??? // How to tell `combine` i==j is not supported?
    }
    Action {
       ... // call swap(i, j)
    }
}
```

## Suggested Solution

Add `combineNotNull`, `flatMapNotNull`, etc methods similar to `net.jqwik.api.Arbitrary#filter(java.util.function.Predicate<T>)`, `net.jqwik.api.Arbitrary#filter(java.util.function.Predicate<T>, int maxMisses)`

```kotlin
combineNotNull(ints(), ints()) { i, j ->
    if (i == j) {
        return null
    }
    Action {
       ... // call swap(i, j)
    }
}
```

See https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/filter-not-null.html , https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/map-not-null.html

jlink:
Sounds useful.

The not-null should be easy to simulate:

```kotlin
combine(ints(), ints()) { i, j ->
    if (i == j) {
        return null
    }
    Action {
       ... // call swap(i, j)
    }
}.filter{ a -> a != null }
```

From a readability standpoint I'd rather have something like:

```java
Combinators.combine(ints(), ints())
   .filter( i, j -> i != j)
   .as(i, j -> new MyAction(...);
```

And similar for Kotlin.

Introducing `CombinatorX.asTuple()` together with Kotlin's deconstruction capabilities would get us _almost_ there I think: 

```kotlin
combine(ints(), ints())
   .asTuple()
   .filter { 
      val i, j = it
      return i != j
   }
   .map {
     val i, j = it
     Action {
       ... // call swap(i, j)
     }
   }
```

vlsi:
The issue with `asTuple()` is that it requires mentioning the same variables multiple times which looks odd.

>The not-null should be easy to simulate:

```kotlin
combine(ints(), ints()) { i, j ->
    if (i == j) {
        return null
    }
    Action {
       ... // call swap(i, j)
    }
}.filter{ a -> a != null }
```

[...]

PR #512, opened 2023-08-03
Kotlin Combinator DSL

fxshlein:
## Overview

Currently, we have a lot of combinators in our kotlin code. They generally look like this:

```kt
fun current(): Arbitrary<String> {
    return combine(
        String.any(),
        String.any(),
        String.any(),
        String.any(),
        String.any(),
        String.any(),
        String.any(),
    ) { first, second, third, fourth, fifth, sixth, seventh ->
        "1: $first, 2: $second, 3: $third, 4: $fourth, 5: $fifth, 6: $sixth, 7: $seventh"
    }
}
```

This has two major disadvantages in my opinion:
- It is hard to read and modify, because you don't directly know which arbitrary belongs to which value
- This only works with up to eight Arbitraries, because of the way the combinator functions have to be written

The syntax that is added by this PR allows to write a combinator like this:

```kt
fun proposed(): Arbitrary<String> {
    return combine {
        val first by String.any()
        val second by String.any()
        val third by String.any()
        val fourth by String.any()
        val fifth by String.any()
        val sixth by String.any()
        val seventh by String.any()

        createAs {
            "1: $first, 2: $second, 3: $third, 4: $fourth, 5: $fifth, 6: $sixth, 7: $seventh"
        }
    }
}
```

### Details

I just wanted to create a PR in case you like this DSL, because I think it would be a cool addition. The actual implementation is somewhat hacky, so I can imagine that it might not be something you want. Otherwise I'll just keep it in our fork we use internally 😉

---

### DIFF
--- a/documentation/src/docs/include/kotlin-module.md
+++ b/documentation/src/docs/include/kotlin-module.md
@@ -446,14 +446,14 @@ combine {
     val second by Arbitraries.strings()
     // ...
 
-    createAs {
+    combineAs {
         "first: $first, second: $second"
     }
 }
 ```
 
 Note that accessing the `first` or `second` properties in the example above 
-_outside_ the `createAs` block would result in an error.
+_outside_ the `combineAs` block would result in an error.
 
 In the background, this is equivalent to:
 
@@ -473,7 +473,7 @@ combine {
     filter { first.isNotEmpty() }
     filter { first != second }
 
-    createAs {
+    combineAs {
         // 'first' will never be empty or equal to 'second'
 
         "first: $first, second: $second"
--- a/documentation/src/docs/include/kotlin-module.md
+++ b/documentation/src/docs/include/kotlin-module.md
@@ -27,6 +27,7 @@ __Table of contents:__
   - [Jqwik Tuples in Kotlin](#jqwik-tuples-in-kotlin)
   - [Type-based Arbitraries](#type-based-arbitraries)
   - [Diverse Convenience Functions](#diverse-convenience-functions)
+  - [Combinator DSL](#combinator-dsl)
 - [Quirks and Bugs](#quirks-and-bugs)
 
 #### Build Configuration for Kotlin
@@ -477,6 +478,20 @@ combine {
         // 'first' will never be empty or equal to 'second'
 
         "first: $first, second: $second"
+    }
+}
+```
+
+And you can also achieve the `flatCombine` behaviour, by using `flatCombineAs`:
+
+```kotlin
+combine {
+    val first by Arbitraries.strings()
+    val second by Arbitraries.strings()
+    // ...
+
+    flatCombineAs {
+        Arbitraries.just("first: $first, second: $second")
     }
 }
 ```
--- a/documentation/src/docs/include/kotlin-module.md
+++ b/documentation/src/docs/include/kotlin-module.md
@@ -463,6 +463,24 @@ combine(listOf(Arbitraries.strings(), Arbitraries.strings())) { values ->
 }
 ```
 
+It is also possible to use filters within the combinator DSL:
+
+```kotlin
+combine {
+    val first by Arbitraries.strings()
+    val second by Arbitraries.strings()
+
+    filter { first.isNotEmpty() }
+    filter { first != second }
+
+    createAs {
+        // 'first' will never be empty or equal to 'second'
+
+        "first: $first, second: $second"
+    }
+}
+```
+
 #### Quirks and Bugs
 
 - Despite our best effort to enrich jqwik's Java API with nullability information,
--- a/documentation/src/docs/include/kotlin-module.md
+++ b/documentation/src/docs/include/kotlin-module.md
@@ -432,6 +432,36 @@ There's a more Kotlinish way to do the same: `anyForType<MyType>()`.
 - `Collection<T>.anySubset()` can replace 
   `Arbitraries.subsetOf(collection: Collection<T>)`
 
+##### Combinator DSL
+
+The combinator DSL provides another, potentially more convenient, wrapper for
+`Combinators.combine`.
+
+Instead of a list of arguments, it uses kotlin's property delegates to refer to
+the arbitraries that are being combined:
+
+```kotlin
+combine {
+    val first by Arbitraries.strings()
+    val second by Arbitraries.strings()
+    // ...
+
+    createAs {
+        "first: $first, second: $second"
+    }
+}
+```
+
+Note that accessing the `first` or `second` properties in the example above 
+_outside_ the `createAs` block would result in an error.
+
+In the background, this is equivalent to:
+
+```kt
+combine(listOf(Arbitraries.strings(), Arbitraries.strings())) { values ->
+    "first: ${values[0]}, second: ${values[1]}"
+}
+```
 
 #### Quirks and Bugs
 
--- a/kotlin/src/main/kotlin/net/jqwik/kotlin/api/CombinatorDsl.kt
+++ b/kotlin/src/main/kotlin/net/jqwik/kotlin/api/CombinatorDsl.kt
@@ -152,7 +152,7 @@ internal class ValueBindings {
 
             block()
         } finally {
-            current.remove()
+            current.set(null)
         }
     }
 }
--- a/kotlin/src/main/kotlin/net/jqwik/kotlin/api/CombinatorDsl.kt
+++ b/kotlin/src/main/kotlin/net/jqwik/kotlin/api/CombinatorDsl.kt
@@ -19,7 +19,7 @@ import kotlin.reflect.KProperty
  *     filter { first.isNotEmpty() }
  *     filter { first != second }
  *
- *     createAs {
+ *     combineAs {
  *         "first: $first, second: $second"
  *     }
  * }
@@ -78,8 +78,8 @@ class CombinatorScope internal constructor(private val bindings: ValueBindings) 
 
 @API(status = API.Status.EXPERIMENTAL, since = "1.8.0")
 sealed class Combined<R>(
-    val arbitraries: List<Arbitrary<*>>,
-    val filters: List<() -> Boolean>
+    private val arbitraries: List<Arbitrary<*>>,
+    private val filters: List<() -> Boolean>
 ) {
     internal fun createArbitrary(bindings: ValueBindings): Arbitrary<R> {
         @Suppress("UNCHECKED_CAST")
--- a/kotlin/src/main/kotlin/net/jqwik/kotlin/api/CombinatorDsl.kt
+++ b/kotlin/src/main/kotlin/net/jqwik/kotlin/api/CombinatorDsl.kt
@@ -1,11 +1,12 @@
 package net.jqwik.kotlin.api
 
 import net.jqwik.api.Arbitrary
+import net.jqwik.api.Combinators
+import net.jqwik.api.Combinators.ListCombinator
 import org.apiguardian.api.API
-import java.util.Optional
+import java.util.concurrent.atomic.AtomicBoolean
 import kotlin.properties.ReadOnlyProperty
 import kotlin.reflect.KProperty
-
 
 /**
  * Combine arbitraries using the combinator DSL:
@@ -14,6 +15,9 @@ import kotlin.reflect.KProperty
  * combine {
  *     val first by Arbitraries.strings()
  *     val second by Arbitraries.strings()
+ *
+ *     filter { first.isNotEmpty() }
+ *     filter { first != second }
  *
  *     createAs {
  *         "first: $first, second: $second"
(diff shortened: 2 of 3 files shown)


vlsi:
This might be relevant as well: https://github.com/jqwik-team/jqwik/issues/329#issuecomment-1273322613

fxshlein:
> This might be relevant as well: [#329 (comment)](https://github.com/jqwik-team/jqwik/issues/329#issuecomment-1273322613)

Good point, we don't really use filtering that much since we try to just generate the arguments, but I can imagine something like this would work within the DSL:
```kt
fun proposed(): Arbitrary<String> {
    return combine {
        val first by String.any()
        val second by String.any()

       filter { 'a' !in first }
       filter { 'b' !in second }
       filter { first != second }
  
        createAs {
            "1: $first, 2: $second"
        }
    }
}
```

Or perhaps, instead of `filter` something more "expressive" like `where` or `require`?

kotlin/src/main/kotlin/net/jqwik/kotlin/api/CombinatorDsl.kt
```diff
@@ -0,0 +1,97 @@
...
+            for (it in combined.properties) {
+                it.unbind()
+            }
+        }
+    }
+}
+
+@API(status = API.Status.EXPERIMENTAL, since = "1.8.0")
+class ArbitraryProperty<T> internal constructor() : ReadOnlyProperty<Nothing?, T> {
+    private val currentValue = ThreadLocal<Optional<T & Any>>()
```
vlsi, line 50: Could you clarify why `ThreadLocal` is needed here? Why isn't a regular reference enough?

vlsi: I think I understand now, however, it would be great to put it in the comment

kotlin/src/main/kotlin/net/jqwik/kotlin/api/CombinatorDsl.kt
```diff
@@ -0,0 +1,97 @@
...
+    val combined = CombinatorScope().combinator()
+
+    @Suppress("UNCHECKED_CAST")
+    return combine(combined.arbitraries as List<Arbitrary<Any?>>) { values ->
+        try {
+            for ((property, value) in combined.properties zip values) {
+                property.bind(value)
+            }
+
+            CombinatorCreateScope().(combined.combinator)()
```
vlsi, line 39: I wonder what are the overheads of the approach. Basicaly you have to create and discard several thread locals for every "combine" call.

I wonder if you can use a single thread-local to pass `list` of the values and access items by the corresponding index. Each property can remember its index, and their get could be `getListOfBoundValues.get(index)`.

WDYT?

kotlin/src/main/kotlin/net/jqwik/kotlin/api/CombinatorDsl.kt
fxshlein: I was assuming here that there could be a feature like parallel execution (now or in the future) that calls the same Arbitrary multiple times in parallel. If that's not the case, then `ThreadLocal` is not necessary of course.

The issue here is that the block passed to `combine` is only called once, but the block passed to `createAs` is called every time the arbitrary has to provide a value. Since the `ArbitraryProperty` is instantiated in `combine`, the instance is shared across all invocations.

This is the hacky part I was talking about 😄

kotlin/src/main/kotlin/net/jqwik/kotlin/api/CombinatorDsl.kt
vlsi: I guess you add `CombinatorCreateScope()` for "improvements in the future", so it might make sense to cache a single object for now to avoid dummy allocations on every call

vlsi:
>Or perhaps, instead of filter something more "expressive" like where or require?

Yeah. I don't ask for including "filter / where" into the current PR, however, it would be nice if the concept could be enhanced to cover more cases in the future.

There's `flatCombine` as well which works as follows (`createAs` returns `Arbitrary`)

```kotlin
fun <T1, T2, R> flatCombine(
    a1: Arbitrary<T1>,
    a2: Arbitrary<T2>,
    flatCombinator: (v1: T1, v2: T2) -> Arbitrary<R>
): Arbitrary<R>
```

kotlin/src/main/kotlin/net/jqwik/kotlin/api/CombinatorDsl.kt
fxshlein: I originally had a global `ThreadLocal`, but I could potentially see issues with some (admittedly very weird) nested calls of `combine`:

```kotlin
fun outer(): Arbitrary<String> {
    return combine {
        val value by Arbitraries.strings()

        createAs {
            value + nested { nestedValue -> nestedValue + value } .sample()
                                                       // ^ the get call for this 
                                                       //   value would now refer
                                                       //   to the list of values
                                                       //   from nested
        }
    }
}

fun nested(mapper: (String) -> String): Arbitrary<String> {
    return combine {
        val value by Arbitraries.strings()

        createAs { mapper(value) }
    }
}
```

Of course this example is absolutely ridiculous, but I think a _global_ `ThreadLocal` could cause some very weird issues.

What should be possible is creating a single `ThreadLocal` inside `combine`, passing it to all properties that are created. Because then, the `ThreadLocal` wouldn't ever be overwritten. Or is that what you are referring to?

kotlin/src/main/kotlin/net/jqwik/kotlin/api/CombinatorDsl.kt
```diff
@@ -0,0 +1,97 @@
...
+
+        arbitraries += this
+        properties += arbitraryProperty
+
+        return arbitraryProperty
+    }
+
+    fun <R> createAs(creator: CombinatorCreateScope.() -> R): Combined<R> {
+        return Combined(arbitraries.toList(), properties.toList(), creator)
+    }
```
vlsi, line 86: I wonder if there should be a validation that `createAs` is called exactly once within the block.
I see you require user to put `createAs` as the final statement, however, what do you think of making `createAs` `return Unit` function, and remember `Combined` within `CombinatorScope` instead?

fxshlein: I think we have to return combined to get kotlin to correctly resolve the return type. But perhaps `@BuilderInference` could also solve this, I'll try it out, one minute

kotlin/src/main/kotlin/net/jqwik/kotlin/api/CombinatorDsl.kt
vlsi: What I mean is "every `combine { ... }` uses its own `TheradLocal` for the list of values.
Of course, nested `combine { ... combine { ... }  }` calls would get two independent thread locals.

kotlin/src/main/kotlin/net/jqwik/kotlin/api/CombinatorDsl.kt
fxshlein: Oh, seems like it works completely without returning `Combined` or `@BuilderInference`. I still like the idea of returning something, because it forces the user to call `createAs` at least once, but adding an error when its called multiple times seems reasonable. What do you think about keeping the return value, and only adding a validation that checks for multiple calls? Then no call would be a compile time error, and multiple calls a runtime error.

fxshlein:
@vlsi I pushed a bunch of changes/suggestions:
- Thread local is now created once per `combine` call
- Added `filter` and `flatCombine` (since it seemed pretty simple to add)
- Added an exception that is thrown when calling `createAs` multiple times

kotlin/src/main/kotlin/net/jqwik/kotlin/api/CombinatorDsl.kt
fxshlein: > it would be great to put it in the comment

Completely missed this comment -  my internet is being kind of flaky right now, so I'm just not receiving comments sometimes 😄

I added some documentation to `ValueBindings` (where the `ThreadLocal` is locade now) 🙂

fxshlein:
I also removed the `CombinatorCreateScope`, I realized it wasn't actually needed, it was from an approach I experimented with before

jlink:
It's a clever solution. Since I'm not doing a lot of Kotlin myself I don't feel ready to judge if the slight improvement of readability (my subjective feeling) is worth introducing yet another combinator DSL; there are already three: `Combinators.combine(..)`, `Builders` and Kotlin extension functions `combine(..)`.

It comes with a performance hit:

```kotlin
@Example
fun `measure performance`() {
    val combinedNew = combine {
        val v1 by Int.any(1..1000)
        val v2 by Int.any(1..1000)

        createAs {
            v1!! + v2!!
        }
    }

    val repetitions: Long = 10_000_000

    val timeNew = measureTime {
        combinedNew.sampleStream().limit(repetitions).forEach { it }
    }
    println("new approach: $timeNew")

    val combinedOld = combine(Int.any(1..1000), Int.any(1..1000)) { v1, v2 -> v1 + v2 }
    val timeOld = measureTime {
        combinedOld.sampleStream().limit(repetitions).forEach { it }
    }
    println("old approach: $timeOld")
}
```

```
new approach: 3.008442949s
old approach: 1.828261633s
```

All code is a liability ;-/

kotlin/src/main/kotlin/net/jqwik/kotlin/api/CombinatorDsl.kt
```diff
@@ -0,0 +1,162 @@
...
+
+/**
+ * A class that provides the [ArbitraryProperty] instances access to the relevant values when accessed from within the
+ * block passed to [CombinatorScope.createAs].
+ */
+internal class ValueBindings {
+    // Because the ArbitraryProperty instances have to be created once per "combine" call (so once per Arbitrary
+    // instance), they are shared across all invocations of the "createAs" block. To prevent issues with potential
+    // parallel generation of arbitrary values, we use a ThreadLocal here.
+    private val current = ThreadLocal<List<*>>()
```
jlink, line 142: Couldn't you get rid of the thread local by introducing a context to `block` in the `withValues` function? Just guessing here since my Kotlin foo is not too developed.

kotlin/src/test/kotlin/net/jqwik/kotlin/CombinatorDslTests.kt
```diff
@@ -0,0 +1,109 @@
...
+            val v2 by just(2)
+
+            createAs {
+                just(v1!! + v2!!)
+            }
+        }
+
+        assertThat(combined.sample()).isEqualTo(3)
+    }
+}
```
jlink, line 109: I'd like at least one test to check if shrinking still works since there's a lot of Kotlin magic going on.

kotlin/src/main/kotlin/net/jqwik/kotlin/api/CombinatorDsl.kt
```diff
@@ -0,0 +1,162 @@
...
+class ArbitraryProperty<T> internal constructor(
+    private val bindings: ValueBindings,
+    private val index: Int,
+) : ReadOnlyProperty<Nothing?, T> {
+    override fun getValue(thisRef: Nothing?, property: KProperty<*>): T {
+        return bindings[index]
+    }
+}
+
+@API(status = API.Status.EXPERIMENTAL, since = "1.8.0")
```
jlink, line 89: This could be `Status.INTERNAL` I assume since it's never called directly by a user.

kotlin/src/main/kotlin/net/jqwik/kotlin/api/CombinatorDsl.kt
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    fun <R> createAs(creator: () -> R): Combined<R> {
+        check(!created.getAndSet(true)) { "'createAs' must only be called once" }
+
+        return Combined(arbitraries.toList(), filters.toList(), creator)
+    }
+}
+
+@API(status = API.Status.EXPERIMENTAL, since = "1.8.0")
```
jlink, line 112: Also `INTERNAL`.

kotlin/src/main/kotlin/net/jqwik/kotlin/api/CombinatorDsl.kt
vlsi: `CombinatorScope` appears in parameter type of user-facing `combine` and `flatCombine`, so it can't be `INTERNAL`.

kotlin/src/main/kotlin/net/jqwik/kotlin/api/CombinatorDsl.kt
```diff
@@ -0,0 +1,162 @@
...
+ *     }
+ * }
+ * ```
+ *
+ * @return new Arbitrary instance
+ */
+@API(status = API.Status.EXPERIMENTAL, since = "1.8.0")
+fun <R> flatCombine(
+    combinator: CombinatorScope.() -> Combined<Arbitrary<R>>
+): Arbitrary<R> {
```
vlsi, line 67: Have you considered using a single `combine` block, and use `combineAs` and `flatCombineAs` factories?
It does not sound right to me to have several top-level `combine` and `flatCombine` builders.

vlsi:
Regarding `ThreadLocal` removal. It looks like delegated properties can't access extra context at the use-site, and I guess the issue is known as https://youtrack.jetbrains.com/issue/KT-10468/Context-receivers-multiple-receivers-on-extension-functions-properties and https://youtrack.jetbrains.com/issue/KT-14504/Support-delegated-properties-with-second-receiver-so-that-they-are-available-only-in-a-certain-context-i.e.-STM .

So the workaround could be adding an extra `.value` which will exist in `BindScope` only:

```kotlin
fun proposed(): Arbitrary<String> {
    return combine {
        val first by String.any()
        val second by String.any()
        val third by String.any()
        val fourth by String.any()
        val fifth by String.any()
        val sixth by String.any()
        val seventh by String.any()

        createAs { // this: BindScope
             // Imagine class BindScope { val PropType.value:  ...
            "1: ${first.value}, 2: ${second.value}, 3: ${third.value}, 4..."
        }
    }
}
```

The explicit `.value` would effectively capture `BindScope`, so `ThreadLocal` would not be needed for passing the values.

An alternative option could be adding a function in `BindScope` context:
For instance:

```kotlin
        createAs { // this: BindScope
             // Imagine class BindScope { operator fun PropType.invoke() =  ...
            "1: ${first()}, 2: ${second()}, 3: ${third()}, 4..."
        }
```

WDYT?

The approach with extra function would likely be Java-friendly as well.

Java syntax could be

```java
var combined = combine((combineScope) -> {
    var first = combineScope.registerVariable(Arbitraries.strings());
    var second = combineScope.registerVariable(Arbitraries.ints());
    combineScope.combineAs((bindScope) ->
        first.getValue(bindScope) + second.getValue(bindScope)
        // or bindScope.get(first) + bindScope.get(second) if it looks better
    });
});
```

In that case, `ThreadLocals` are not needed, and the same API could more-or-less work for both Java and Kotlin.

Of course, as Kotlin implements context receivers, it might provide a nicer API without explicit `.value`.

WDYT?

jlink:
> An alternative option could be adding a function in BindScope context:
> For instance:
```
createAs { // this: BindScope
             // Imagine class BindScope { operator fun PropType.invoke() =  ...
            "1: ${first()}, 2: ${second()}, 3: ${third()}, 4..."
        }
```

Couldn't the property's getter function bind to the scope?
I'm still hoping there's a way to stick with property access without using a thread local.

jlink:
> The approach with extra function would likely be Java-friendly as well.

That has no importance to me. Feels completely unnatural in Java.

fxshlein:
> Couldn't the property's getter function bind to the scope?

I've been experimenting with this, but the issue is that kotlin takes whatever `getValue` function it finds at the place where the property is delegated (so essentially wherever the `by` keyword is) and then runs with that, without a way to overwrite that.

kotlin/src/main/kotlin/net/jqwik/kotlin/api/CombinatorDsl.kt
```diff
@@ -0,0 +1,162 @@
...
+ */
+internal class ValueBindings {
+    // Because the ArbitraryProperty instances have to be created once per "combine" call (so once per Arbitrary
+    // instance), they are shared across all invocations of the "createAs" block. To prevent issues with potential
+    // parallel generation of arbitrary values, we use a ThreadLocal here.
+    private val current = ThreadLocal<List<*>>()
+
+    operator fun <T> get(index: Int): T {
+        val current = checkNotNull(current.get()) {
+            "Arbitrary delegate property must only be used inside 'combineAs' or 'filter'"
```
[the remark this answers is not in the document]
fxshlein: Renamed everything to `combineAs` and `flatCombineAs` so it fits more use cases, and also matches the other APIs

fxshlein:
Essentially, kotlin needs a `.getValue()` function that works inside `combine`:
```kotlin
        val combined = combine {
            // Kotlin determines which 'ArbitraryProperty<T>.getValue()'
            // function it will use here, so it has to work here:
            val value by just(1)

            combineAs {
                // So even if another ' getValue' is available in this scope, it is ignored.
                value
            }
        }
```

For that reason, you cannot bind it to the scope in `combineAs`, because it will always have to bind to the scope in `combine`.

If there is no `.getValue()` at all available in `combine`, kotlin will complain - so only making it available in `combineAs` is also not possible.

jlink:
> If there is no `.getValue()` at all available in `combine`, kotlin will complain - so only making it available in `combineAs` is also not possible.

I wonder if one of Kotlin's scope functions could make a difference, like:
```kotlin
val combined = combine {
            // Kotlin determines which 'ArbitraryProperty<T>.getValue()'
            // function it will use here, so it has to work here:
            val value by just(1)

            combineAs {
                bindings.let {
                     value // would it use value in bindings?
                }
            }
        }
```

kotlin/src/test/kotlin/net/jqwik/kotlin/CombinatorDslTests.kt
```diff
@@ -106,4 +107,17 @@ class CombinatorDslTests {
 
         assertThat(combined.sample()).isEqualTo(3)
     }
+
+    @Property
```
jlink, line 111: I tend to reduce the number of tries for these kinds of shrinking tests to 10 or 100 max, which makes certain that not only edge cases have been created and shrinking works in general.

fxshlein: Oh oops, I overlooked the `@PropertyDefaults` in the class I blatantly copied from, my bad

fxshlein:
> > If there is no `.getValue()` at all available in `combine`, kotlin will complain - so only making it available in `combineAs` is also not possible.
> 
> I wonder if one of Kotlin's scope functions could make a difference, like:

Sadly not, because they also just use scopes and receivers in the same way we can use them.

fxshlein:
I think there is no way around having either the `ThreadLocal`, or not having the property access syntax.

I profiled a bit btw, and a lot of the performance drop actually comes from the `ThreadLocal.remove()`?! I wonder if it would be fine and speed stuff up to just replace that with `set(null)`

kotlin/src/main/kotlin/net/jqwik/kotlin/api/CombinatorDsl.kt
```diff
@@ -0,0 +1,158 @@
...
+    private val index: Int,
+) : ReadOnlyProperty<Nothing?, T> {
+    override fun getValue(thisRef: Nothing?, property: KProperty<*>): T {
+        return bindings[index]
+    }
+}
+
+@API(status = API.Status.EXPERIMENTAL, since = "1.8.0")
+class CombinatorScope internal constructor(private val bindings: ValueBindings) {
+    private var created = AtomicBoolean(false)
```
vlsi, line 52:
```suggestion
    private val created = AtomicBoolean(false)
```

fxshlein:
I've done some more benchmarking. The test you sent above on my machine:
```
new approach: 2.482132500s
old approach: 1.483676100s
```

However, replacing `ThreadLocal.remove()` with `.set(null)`, I get this:
```
new approach: 1.942857400s
old approach: 1.645651901s
```

That's quite the big difference, but I think I figured out why: Since ThreadLocal.remove()` will completely remove the value for that Thread from the map of thread local values, the next call to `.set(values)` will have to make a completely new entry in the map. If I only call `.set(null)` however, I still remove the reference to the values so they can be garbage collected, but the next `.set(values)` call will only have to update the existing entry that is still set to null, instead of having to create a completely new entry. So it makes it more performant on both ends.

fxshlein:
I think in this case, `.set(null)` is actually fine. The `ThreadLocalMap` uses weak references to keep the `ThreadLocal` around, so it would essentially just be garbage collected together with the `Arbitrary`, even without ever calling `.remove()`.

Since the whole deal of Arbitraries is basically doing a lot of repetition, and probably on the same thread, I would argue to use `.set(null)` to clear the `ThreadLocal` here for the performance benefit.

documentation/src/docs/include/kotlin-module.md
```diff
@@ -432,6 +432,54 @@ There's a more Kotlinish way to do the same: `anyForType<MyType>()`.
 - `Collection<T>.anySubset()` can replace 
   `Arbitraries.subsetOf(collection: Collection<T>)`
 
+##### Combinator DSL
```
jlink, line 435: Please add `Combinator DSL` to this file's TOC.

jlink:
@fxshlein Do you consider the PR ready for merging?

fxshlein:
> @fxshlein Do you consider the PR ready for merging?

I think so yeah - the `ThreadLocal` is unfortunate, but I still can't think of any way around it 😄

jlink:
> the `ThreadLocal` is unfortunate, but I still can't think of any way around it 😄

Your `ThreadLocal` is in good company in jqwik's source.

fxshlein:
Cool! Thank you for merging (And of course for jqwik - it's a really cool tool, and I really enjoy writing tests with it!)

> Your `ThreadLocal` is in good company in jqwik's source.

Haha, perhaps... I've actually seen quite a lot of the source code - My first experiment in our jqwik fork was a lot more... intrusive I'd say 😄
You can see the diff here: https://github.com/jqwik-team/jqwik/compare/main...Dyescape:jqwik:feature/dynamic-parameters

I actually got it to work like that including shrinking and everything, but it's probably not completely possible to implement it like this sadly. One day I'll probably come back to a similar Idea tho.

jlink:
> I actually got it to work like that including shrinking and everything, but it's probably not completely possible to implement it like this sadly. One day I'll probably come back to a similar Idea tho.

Cool idea. I guess a proper implementation would also require a better separation of concerns in jqwik's engine code base.

Issue #523, opened 2023-09-26
Consider supporting @UseType for sealed interfaces


dlesl:
## Testing Problem

Here is an example of testing a `sealed interface` which represents a data type.

```java
class ExampleProperty {
  sealed interface JsonValue {
    record Text(String text) implements JsonValue {}
    record Number(double value) implements JsonValue {}
  }

  @Property
  @Domain(DomainContext.Global.class)
  @Domain(JsonDomain.class)
  void example(@ForAll @UseType JsonValue jsonValue) {
    System.out.println(jsonValue);
  }

  static class JsonDomain extends DomainContextBase {
    @Provide
    Arbitrary<JsonValue> jsonValue() {
      return Arbitraries.of(JsonValue.class.getPermittedSubclasses())
          .flatMap(c -> Arbitraries.forType(c).map(j -> (JsonValue) j));
    }
  }
}

```

Perhaps this strategy could be attempted by jqwik's `@UseType` by default?

## Discussion

The current behaviour is that generation fails immediately for a sealed interface. Having this as the default behaviour could potentially be worse if it fails to discover/generate some of the subclasses and they get missed in testing?

jlink:
Sounds like a very reasonable idea.

Currently I've no build setup that allows me to use Java 17 features.
I wonder if an additional module for Java 17 would do.

jibidus:
Same for kotlin module would be great.

oleschoenburg:
Just FYI, If the type hierarchy is deeper, recursive search for subclasses is needed: 
```java
  public static <T> Stream<Class<T>> implementationsOfSealedInterface(final Class<T> clazz) {
    if (!clazz.isSealed()) {
      throw new IllegalArgumentException(String.format("Class %s is not sealed", clazz.getName()));
    }
    return Stream.of(clazz.getPermittedSubclasses())
        .flatMap(
            c -> {
              if (c.isSealed()) {
                return implementationsOfSealedInterface((Class<T>) c);
              } else {
                return Stream.of((Class<T>) c);
              }
            });
  }
```

jlink:
Since work for Jqwik2 will probably be started in the upcoming months, this feature might have to wait till then.

jlink:
Kotlin only

> Another solution with Kotlin where we use custom arbitrary for some subtypes:
> 
> Usage:
> 
> ```kotlin
> anyForSubtype<MyInterface> {
>    use<MySubtype> { mySubTypeArbitrary() }
> }
> ```

[...]

PR #555, opened 2024-02-26
Add anyForType() dedicated to sealed classes and interfaces in Kotlin module: anyForSubtypeOf()

jibidus:
## Overview

`anyForType()` does not support sealed classes or interfaces. The idea here is to support these cases with a dedicated function `anyForSubtypeOf()` for Kotlin only, since it is easier to implement than in Java.

### Details

`anyForSubtypeOf()` use `TypeArray` under the hood for each sealed subtypes, recursively:

```kotlin
sealed interface Character
sealed interface Hero : Character
class Knight(val name: String) : Hero
class Wizard(val name: String) : Character

val arbitrary =  anyForSubtypeOf<Character>(enableArbitraryRecursion = true) {
    provide<Wizard> { Arbitraries.of(Wizard("Merlin"),Wizard("Élias de Kelliwic’h")) }
}
```

It is also possible to enable recursion with created `TypeArray` with `enableArbitraryRecursion`, or to provide custom arbitraries for specific subtypes.

### Notable decisions

* `enableRecursion` is not handled through a fluent api (but with a parameter), which seems to be the common way to do such things in jqwik. I didn't succeed to implement this feature with a dedicated type (ex `SubtypeArbitrary`), which would permit to use a fluent api.
* since many concepts are created here, I used a dedicated file `AnyForSubtypeOfDsl.kt` (instead of `JqwikGlobals.kt` where `anyForType()` is declared).
* I would have liked to group tests with their test cases, in order to reuse same interface en class names for readability purpose. But I didn't succeed to use inner classes annotated with `@Group` annotations, since sealed interface cannot be declared in inner classes.
* It would probably be more intuitive for jqwik users, to add this feature in `anyForType()`, but I'm not familiar enough with `DefaultTypeArbitrary` and `TraverseArbitrary` classes to be able to do that. Maybe with a little help…

---


jibidus:
This will fix #523 for Kotlin only.

jibidus:
I see many difficulties with this simple `if` :
* this `if` should be in jqwik engine, where `isSealed()` is not available (not in kotlin module.
* it's not clear for me where to put this `if`. Maybe in `DefaultTraverseArbitrary`?

jlink (commented):
I think the entry point for `anyForSubtypeOf(..)` should be in file `JqwikGlobals.kt`, next to `anyForType`. Of course, you can delegate from there to the real implementation in a different file.

jlink:
> unless you were thinking of putting this `if` in `anyForType()` function? It would be better for jqwik users, but `anyForSubtype()` have to return an instance of `TypeArbitrary`, and this is one of the difficulties I have had to face.

I think we can deal with this later. Let's get `anyForSubtype()` ready for release first.

jlink:
> * this `if` should be in jqwik engine, where `isSealed()` is not available (not in kotlin module.

Could
https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.reflect/-k-class/is-sealed.html do the trick?

kotlin/src/main/kotlin/net/jqwik/kotlin/api/JqwikGlobals.kt
```diff
@@ -106,6 +106,70 @@ inline fun <reified T> anyForType(): TypeArbitrary<T>
...
+ * If a subtype is a sealed class or interface, its subtypes are used to create [Arbitrary]. This is done recursively.
+ * [TypeArbitrary] are created by default under the hood, but this can be customized, for each subtype, with [SubtypeScope.provide] .
+ * ```kotlin
+ * anyForSubtypeOf<MyInterface> {
+ *     provide<MyImplementation1> { customArbitrary1() }
+ *     provide<MyImplementation2> { customArbitrary2() }
+ * }
+ * ```
+ * @param enableArbitraryRecursion is applied to all created [TypeArbitrary].
+ */
```
jlink, line 121: Add api annotation: 
`@API(status = API.Status.EXPERIMENTAL, since = "1.8.4")`

kotlin/src/main/kotlin/net/jqwik/kotlin/api/JqwikGlobals.kt
```diff
@@ -106,6 +106,70 @@ inline fun <reified T> anyForType(): TypeArbitrary<T>
...
+inline fun <reified T> anyForSubtypeOf(
+    enableArbitraryRecursion: Boolean = false,
+    crossinline subtypeScope: SubtypeScope<T>.() -> Unit = {}
+): Arbitrary<T> where T : Any {
```
jlink, line 125: Would it be possible to have `TypeArbitrary<T>` as return type?

jlink (commented):
Since the implementation is sufficiently complex I'd rather have it in a separate file and the function in `JqwikGlobals` delegating to it. What do you think?

kotlin/src/main/kotlin/net/jqwik/kotlin/api/JqwikGlobals.kt
jibidus: This would imply that custom arbitrary (registered with `SubtypeScope<T>.provide()`) can only be `TypeArbitrary<T>`. This would make custom arbitraries useless, wouldn't it?

jibidus:
> Since the implementation is sufficiently complex I'd rather have it in a separate file and the function in `JqwikGlobals` delegating to it. What do you think?

I have no opinion.

With only `anyForSubtypeOf()` in `JqwikGlobals`, we follow a technical cohesion (with only entry points gathered in `JqwikGlobals`). On the other side, with all components of `anyForSubtypeOf` feature in the same place, we follow a functional cohesion.
It seems that the first one permits to follow convention of current codebase (which is great), whereas the second one permits to follow a better cohesion (based on functionalities, not on technical characteristics).

So I have no opinion.
I have done what you suggest 🙂.

PR #576, opened 2024-06-06
Improve nullability annotations and generic variance

vlsi:
This is a draft, however, it should more-or-less work.

Notable changes:
* `@NullMarked` package annotation does not automatically inherit to sub-packages, so `@NullMarked` must be duplicated manually (I've added many `package-info.java` files)
* By default, Kotlin assumes that type parameter is nullable. For instance `<T> fun(value: T) {...` means that it is valid to use `String?` for `T`. At the same time, Java defaults to non-null, so if nullable T should be allowed, then the declaration should be `<T extends @Nullable Object>`
* In 99.9% of the cases `@NotNull` is not needed. If IDEA suggest "overriding method does ..." and suggests adding `@NotNull`, then most likely `@NullMarked` `package-info.java` is missing
* I updated `Consumer<T>` -> `Consumer<? super T>`; `Function<ARG, RESULT>` -> `Function<? super ARG, ? extends RESULT>` and so on. Unfortuantely, with Java one has to duplicate those super/extends all over the place
* In most cases `List<? extends T>` is effectively a `ReadOnlyList<T>`. It is helpful in both generic variance (accepting more matching arguments) and it helps to understand the method signature as in "ok, the method accepts `List<? extends T>`, so it does not modify the list"


lcahmu:
Thanks @vlsi, very much appreciated! I tried using this branch in my project and get a Checker error using `Arbitrary.just(..)`. Looks like that one got overlooked?

vlsi:
What is the error?

lcahmu:
It's a type mismatch. It's still `<@Nullable T> Arbitrary<T> just(@Nullable T)` where nearby methods were updated to `<T extends @Nullable Object>`. `ArbitrariesFacade.just(..)` was updated that way, so looks like it just got missed on `Arbitrary`.

jlink:
@vlsi What's missing in your opinion to make it a mergeable, non-WIP PR?

api/src/main/java/net/jqwik/api/Arbitraries.java
```diff
@@ -454,7 +454,7 @@ public static CharacterArbitrary chars() {
 	 * @return a new arbitrary instance
 	 */
 	@API(status = MAINTAINED, since = "1.1.1")
-	public static <T> Arbitrary<T> create(Supplier<T> supplier) {
```
jlink, line 457: I'm not sure that methods that previously returned <T> should now return <T extends @Nullable...>. It might be the right choice, but then it should be a different PR IMO.

api/src/main/java/net/jqwik/api/Builders.java
```diff
@@ -217,7 +217,7 @@ public BuilderCombinator<B> in(BiFunction<B, @Nullable T, B> toFunction) {
 		 * @param setter Use value provided by arbitrary to change a builder's property.
 		 * @return new {@linkplain BuilderCombinator} instance with same embedded builder
 		 */
-		public BuilderCombinator<B> inSetter(BiConsumer<B, T> setter) {
```
jlink, line 220: Same argument here. If correct (have to think it through), then a different PR.

api/src/main/java/net/jqwik/api/arbitraries/StreamableArbitrary.java
```diff
@@ -79,6 +79,6 @@ default StreamableArbitrary<T, U> ofSize(int size) {
 	 * @return new arbitrary instance
 	 */
 	@API(status = MAINTAINED, since = "1.7.3")
-	StreamableArbitrary<@Nullable T, U> uniqueElements(Function<@Nullable T, Object> by);
+	StreamableArbitrary<T, U> uniqueElements(Function<? super T, ?> by);
```
jlink, line 82: The change to the second type parameter should then also entail a change in `@UniqueElements` annotation and the classes that process it.

kotlin/src/main/kotlin/net/jqwik/kotlin/api/BuildersExtensions.kt
```diff
@@ -9,10 +9,8 @@ import java.util.function.BiFunction
  * Convenience function for Kotlin to not use backticked `in` function.
  */
 @API(status = API.Status.EXPERIMENTAL, since = "1.6.0")
-fun <B, T> BuilderCombinator<B>.use(arbitrary: Arbitrary<T>, combinator: (B, T) -> B): BuilderCombinator<B>
+fun <B, T: Any?> BuilderCombinator<B>.use(arbitrary: Arbitrary<T>, combinator: (B, T) -> B): BuilderCombinator<B>
     where B : Any {
-    // This explicit conversion form combinator to toFunction is necessary
```
jlink, line 14: The comment is still valid, isn't it?

api/src/main/java/net/jqwik/api/Arbitraries.java
vlsi: Do you mean you want forbid creating `Arbitrary<String?>` with this method?

Consider the use case:  `Arbitraries.create(() -> random.nextBoolean() ? "hi" : null)`.
Previous declaration (`<T>`) forbids such use cases, however, they are pretty valid.

api/src/main/java/net/jqwik/api/arbitraries/StreamableArbitrary.java
vlsi: Why?

jlink: `Class<? extends Function<?, ?>> by()` would allow specificly typed implementations. But it can also be done in a follow-up PR.

kotlin/src/main/kotlin/net/jqwik/kotlin/api/BuildersExtensions.kt
vlsi: It is not. `toFunction` is no longer neded

jlink: It's still being used in your code, though. I'm confused.

vlsi:
I think it should be good to go assuming the tests pass.
I understand the PR somehow includes both nullability and generic variance changes, however often it is the case `@Nullable` is at the same line as `? extends/super`. I would take significant effort to split the changes.

Note: there might be many more generic variance improvements to come, and I touched only the signatures that were close to the nullability fixes.

I'm fine answering questions and resolving the outstanding issues, however I will not spend my time on splitting the changes into different PRs.

kotlin/src/main/kotlin/net/jqwik/kotlin/api/BuildersExtensions.kt
vlsi: I am afraid it is not used. Please provide an exact link where `toFunction` is used in `BuilderExtensions.kt`

jlink: Line 15 in BuilderExtensions.kt reads 
`return this.use(arbitrary).`in`(toFunction)`

vlsi: Please ensure you use the latest code.
Here's my branch: https://github.com/vlsi/jqwik/blob/nullness_annotations/kotlin/src/main/kotlin/net/jqwik/kotlin/api/BuildersExtensions.kt

Line 15 is the final closing curly brace: https://github.com/vlsi/jqwik/blob/abe7b1e99f6656eb26cdeea864f90f03867e4313/kotlin/src/main/kotlin/net/jqwik/kotlin/api/BuildersExtensions.kt#L15

jlink: Now I see the final code. Maybe GH cached the old state or I viewed it before you pushed. Whatever.

jlink:
Many thanks @vlsi for the work and the focus on so many details!

It also show how tedious it can become to implement null safety and variance constraints in a language that has not be designed for that. I'm hesitant if I'll aim for that in jqwik 2, too.

PR #633, opened 2025-05-07
Make `description` method public in `JustTransform`

mourjo:
## Overview

Make `description` method public in `JustTransform` - this will allow the extender of `JustTransform` to use custom descriptions.

### Details
For example, with `JustMutate`, it is possible to override the description:
```java
class RemoveAction extends JustMutate<List<Integer>> {
    
    String description = "RemoveAction";

    @Override
    public void mutate(List<Integer> state) {
        description = "Mutating by removing something";
    }

    @Override
    public boolean precondition(List<Integer> state) {
        return true;
    }

    @Override
    public String description() {
        return description; // returns the latest description
    }
}
```

But it is not possible to do this with `JustTransform`:
```java
class AddAction extends JustTransform<List<Integer>> {

    String description = "RemoveAction";

    @Override
    public List<Integer> transform(List<Integer> state) {
        description = "Mutating by adding something";
        return state;
    }

    @Override
    public boolean precondition(List<Integer> state) {
        return true;
    }

    /*
    // this is not possible:
    @Override
    public String description() {
        return description; // returns the latest description
    }
    */
}
```
---

### DIFF
--- a/api/src/main/java/net/jqwik/api/state/Action.java
+++ b/api/src/main/java/net/jqwik/api/state/Action.java
@@ -89,7 +89,7 @@ public interface Action<S extends @Nullable Object> {
 		 * Return an arbitrary for transformers that depends on the previous state.
 		 *
 		 * <p>
-		 * In addition to performing a state transformation the transformin function
+		 * In addition to performing a state transformation the transforming function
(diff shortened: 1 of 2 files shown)


jlink:
This makes absolute sense and probably was an oversight of mine in the first place.

PR #634, opened 2025-05-11
perf: make anyForSubtypeOf cacheable

### LINKED ISSUE #624: Performances degradataion in 1.9.0
Hi @jlink, 
I noticed a performance degradation since jqwik 1.9.0 in some of my tests.
I have many tests which heavily use **Domain** feature, which duration have increased by a factor 5. For example, here are durations for a single test:

| Jqwik version  | Test duration |
| -------- | ------- |
| 1.8.5      | 7s         |
| 1.9.0      | 36s      |
| 1.9.1      | 36s       |
| 1.9.2      | 36s       |

I will try to provide a project to illustrate this issue.

vlsi:
## Overview

This is a prototype that might resolve https://github.com/jqwik-team/jqwik/issues/624

### Details

It turns out kotlinc inserts "sam wrappers" when converting Kotlin lambdas into Java functional interfaces. The sam wrappers do not have equals/hashCode, so it defeats jqwik's caching.

The fix is to move out `.flatMap { ... }` out of `inline` function, so kotlinc no longer adds [`sam wrappers`](https://github.com/JetBrains/kotlin/blob/54fa2d938aee9fc428ffd9b00ef2791cef3ec430/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/lower/SingleAbstractMethodLowering.kt#L159).

TODO:
- [ ] `DefaultTypeArbitrary.equals` should account for `traverseArbitrary.enableRecursion` somehow
- [ ] `net.jqwik.kotlin.api.SubtypeScope#hashCode` does not seem to trigger when executing `AnyForSubtypeOfTests`. I guess it should somehow
- [ ] I haven't actually compared the performance for `anyForSubtypeOf`

---

### DIFF
--- a/api/src/main/java/net/jqwik/api/Arbitrary.java
+++ b/api/src/main/java/net/jqwik/api/Arbitrary.java
@@ -1,5 +1,6 @@
 package net.jqwik.api;
 
+import java.io.*;
 import java.util.*;
 import java.util.function.*;
 import java.util.stream.*;
--- a/engine/src/main/java/net/jqwik/engine/properties/arbitraries/DefaultTypeArbitrary.java
+++ b/engine/src/main/java/net/jqwik/engine/properties/arbitraries/DefaultTypeArbitrary.java
@@ -4,6 +4,8 @@ import java.lang.reflect.*;
 import java.util.*;
 import java.util.function.*;
 
+import net.jqwik.api.support.*;
+
 import org.jspecify.annotations.*;
 import org.junit.platform.commons.support.*;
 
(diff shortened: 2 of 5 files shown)


jlink:
Currently not working on it. See [Status: Maintenance Mode](https://github.com/jqwik-team/jqwik?tab=readme-ov-file#maintenance-mode)

jlink:
Closing to respect @vlsi's decision that he no longer wants to be involved with jqwik due to anti-AI clause.

PR #636, opened 2025-05-12
fix: use try-with-resources for ObjectOutputStream to ensure all the data is written

vlsi:
Previously, ObjectOutputStream could silently buffer data in the internal buffers

---


jlink:
@vlsi Thanks!

